PR #6, opened 2024-10-28
refactor: move ProtocolHandler impl

dignifiedquire:


src/net_protocol.rs
```diff
@@ -0,0 +1,309 @@
...
+            }
+        }
+        match last_err {
+            Some(err) => Err(err.into()),
+            None => Err(anyhow!("No nodes to download from provided")),
+        }
+    }
+}
+
+impl<S: crate::store::Store> ProtocolHandler for Blobs<S> {
```
rklaehn, line 290: These implementations don't really look beautiful with all the Arc and Box... But 🤷‍♂️ .

rklaehn (approved):
LGTM

PR #7, opened 2024-10-31
feat: add rpc protocol, client and server

rklaehn:
## Description

Add rpc protocol, client and server

## Breaking Changes

- net_protocol::Blobs::new_with_events takes an Endpoint

Compared to the old rpc in iroh, client::blobs::share is gone. That rpc call was always a bit questionable, it just takes a hash and adds the node data, and as such really has nothing to do with blobs as such.

## Notes & open questions

~~Merge this as is, or first try out the rpc refactorings?~~

This is now against the new quic-rpc. There are a number of types in net_protocol that could be made private with some effort, but I would prefer to do this in a subsequent PR. All fns on Blobs have been made private already though.

## Change checklist

- [x] Self-review.
- [x] ~~Tests if relevant.~~


Cargo.toml
```diff
@@ -65,13 +73,18 @@ rcgen = "0.12.0"
...
 
 [features]
-default = ["fs-store"]
+default = ["fs-store", "rpc"]
 downloader = ["dep:parking_lot", "tokio-util/time", "dep:hashlink"]
 fs-store = ["dep:reflink-copy", "redb", "dep:redb_v1", "dep:tempfile"]
 metrics = ["iroh-metrics/metrics"]
 redb = ["dep:redb"]
+rpc = ["dep:quic-rpc", "dep:quic-rpc-derive", "dep:nested_enum_utils", "dep:strum", "dep:futures-util", "dep:ref-cast", "dep:portable-atomic", "dep:walkdir", "downloader"]
+ref-cast = ["dep:ref-cast"]
```
dignifiedquire: why is this its own feature?

Cargo.toml
```diff
@@ -65,13 +73,18 @@ rcgen = "0.12.0"
...
+portable-atomic = ["dep:portable-atomic"]
```
dignifiedquire: why is this its own feature?

Cargo.toml
```diff
@@ -65,13 +73,18 @@ rcgen = "0.12.0"
...
+walkdir = ["dep:walkdir"]
```
dignifiedquire: why is this its own feature?

src/net_protocol.rs
```diff
@@ -149,6 +152,14 @@ impl<S: crate::store::Store> Blobs<S> {
         &self.store
     }
 
+    pub(crate) fn rt(&self) -> LocalPoolHandle {
```
dignifiedquire, line 155: might be useful to make this public, not sure

src/net_protocol.rs
```diff
@@ -149,6 +152,14 @@ impl<S: crate::store::Store> Blobs<S> {
         &self.store
     }
```
dignifiedquire, line 154: I think down in this file there are a bunch of method which are public, only to make RPC work, would be good to check which of those you can now make private

src/rpc.rs
```diff
@@ -0,0 +1,871 @@
...
+    {
+        use Request::*;
+        match msg {
+            Blobs(msg) => self.handle_blobs_request(msg, chan).await,
+            Tags(msg) => self.handle_tags_request(msg, chan).await,
+        }
+    }
+
+    /// Handle a tags request
+    pub async fn handle_tags_request<C>(
```
dignifiedquire, line 76: does this need to be public?

src/rpc/client/blobs.rs
```diff
@@ -0,0 +1,1795 @@
...
+use crate::rpc::proto::blobs::{
+    AddPathRequest, AddStreamRequest, AddStreamUpdate, BatchCreateRequest, BatchCreateResponse,
+    BlobStatusRequest, ConsistencyCheckRequest, CreateCollectionRequest, CreateCollectionResponse,
+    DeleteRequest, ExportRequest, ListIncompleteRequest, ListRequest, ReadAtRequest,
+    ReadAtResponse, ValidateRequest,
+};
+
+/// Iroh blobs client.
+#[derive(Debug, Clone)]
+pub struct Client<C = BoxedConnector<RpcService>> {
```
dignifiedquire, line 109: why is this not repr transparent?

src/rpc.rs
rklaehn: nope

src/rpc/client/blobs.rs
rklaehn: I don't need it in this direction. But you never know... So going to add it.

src/net_protocol.rs
rklaehn: Making all of this private would be a bit of effort.

PR #10, opened 2024-11-13
feat: `import_verifiable_stream` and `write_verifiable_stream` in `Store`

matheus23:
## Description

- Adds a default-implemented fn `Store::import_verifiable_stream`
- Adds a default-implemented fn `MapEntry::write_verifiable_stream`

These allow interacting with the `iroh_blobs` store using iroh's bao-style streams.

---

I'm coming across this while extracting `iroh-willow` into its own thing. This means merging in all work that's not yet merged into iroh yet from the `willow` PR/branch in iroh. This is one of these things.

IIUC, this allows us to stream the willow entry payloads directly from the WGPS protocol instead of having to simultaneously run iroh-blobs & iroh-willow.

## Breaking Changes

Adding fn to traits should be non-breaking as long as they have default impls, right?

## Notes & Open Questions

Unfortunately no tests with this yet, other than my pinky-promise that this code has worked so far for iroh-willow.
That said, I haven't tried bigger payloads yet.

I'll probably make use of this branch for the iroh-willow extraction and come back eventually to add proper tests.

## Change checklist

- [x] Self-review.
- [x] Tests if relevant.


matheus23:
I still requested a review, just to confirm: Is this a feature we want to support in `iroh-blobs`?
If so, I'll come back and add some tests to this.
If not, we'll need to think of some other way of doing this & revise the use in `iroh-willow`.

src/store/traits.rs
```diff
@@ -90,6 +95,31 @@ pub trait MapEntry: std::fmt::Debug + Clone + Send + Sync + 'static {
     fn outboard(&self) -> impl Future<Output = io::Result<impl Outboard>> + Send;
     /// A future that resolves to a reader that can be used to read the data
     fn data_reader(&self) -> impl Future<Output = io::Result<impl AsyncSliceReader>> + Send;
+
+    /// Encodes data and outboard into a [`AsyncStreamWriter`].
+    ///
+    /// Data and outboard parts will be interleaved.
+    ///
+    /// `offset` is the byte offset in the blob to start the stream from. It will be rounded down to
```
rklaehn, line 103: Are you sure this will round down to the next chunk group? The ranges are given in chunks, not chunk groups, so it should round down to the next chunk. So e.g. 1026 would round down to 1024.

In addition: I am sure both sides are compatible with each other, but I am not sure if they are compatible with bao without chunk groups, as specified in WGPS. Do we care?

What bao-tree will do is the following: It will round down to individual chunks, but only where needed. So e.g. if you ask for `31*1024..65*1024`, it will send the chunk from `32*1024..64*1024` as a single chunk of data without any intermediate hashes. If you want to be compatible with bao without chunk groups you would have to send this as 16 individual 1024 byte chunks with intermediate hash pairs.

matheus23: > In addition: I am sure both sides are compatible with each other, but I am not sure if they are compatible with bao without chunk groups, as specified in WGPS. Do we care?

[WGPS allows payload transformations](https://willowprotocol.org/specs/sync/index.html#sync_payloads_transform). So it's just the iroh flavor of WGPS payload transfer.

> Are you sure this will round down to the next chunk group? The ranges are given in chunks, not chunk groups, so it should round down to the next chunk. So e.g. 1026 would round down to 1024.

I need to have a proper look. I've mostly shuffled around the code, I'm just the messenger :)

rklaehn: If we are OK with just being incompatible with anything else, I think it will work.

src/store/traits.rs
```diff
@@ -341,6 +371,74 @@ pub trait Store: ReadableStore + MapMut + std::fmt::Debug {
...
 
+    /// Import a blob from a verified stream, as emitted by [`MapEntry::write_verifiable_stream`];
+    ///
+    /// `total_size` is the total size of the blob as reported by the remote.
+    /// `offset` is the byte offset in the blob where the stream starts. It will be rounded
+    /// to the next chunk group.
+    fn import_verifiable_stream<'a>(
+        &'a self,
+        hash: Hash,
+        total_size: u64,
```
rklaehn, line 382: Where does the total_size come from? Is it a verified size?

rklaehn:
write_verifiable_stream seems identical to https://docs.rs/iroh-blobs/latest/iroh_blobs/provider/fn.send_blob.html except that send_blob is more generic.

rklaehn:
Quick update on this: there are going to be a lot of changes to the store before 1.0. I am not happy with the current trait hierarchy at all. But this functionality will exist in some way in the new api.

rklaehn:
> That's perfect. I'm happy to change iroh-willow to use any eqivalent new abstractions once they're in (doesn't have to be this PR).

Basically I want to get rid of the entire trait hierarchy and replace it with a "protocol" that you use to talk to the store. There will be 2 protocol calls ImportBao and ExportBao, and these will then be wrapped in a pleasant way by an API that sits in front of the protocol.

This is the bao impex part of the protocol:

```rust
/// Export the given sizes in bao format, with the iroh block size.
///
/// The returned stream should be verified by the store.
#[derive(Debug)]
pub struct ExportBao {
    pub hash: Hash,
    pub ranges: ChunkRanges,
    pub out: mpsc::Sender<EncodedItem>,
}

#[derive(Debug)]
pub struct ImportBao {
    pub hash: Hash,
    pub size: u64,
    pub data: mpsc::Receiver<BaoContentItem>,
    pub out: oneshot::Sender<anyhow::Result<()>>,
}
```

And this is how it will be exposed in the "pleasant" API:

```rust
    pub async fn import_bao_quinn(
        &self,
        hash: Hash,
        ranges: ChunkRanges,
        stream: &mut quinn::RecvStream,
    ) -> anyhow::Result<()> { ... }

    pub async fn import_bao_bytes(
        &self,
        hash: Hash,
        ranges: ChunkRanges,
        data: Bytes,
    ) -> anyhow::Result<()> { ... }
```

rklaehn:
Closing. Will keep track of this in #61

PR #16, opened 2024-11-21
refactor: Move the simple node from test scope to behind a feature flag

rklaehn:
## Description

Move the simple node from test scope to behind a feature flag and use it not just from tests but also from examples

## Change checklist

- [ ] Self-review.
- [ ] Tests if relevant.


Cargo.toml
```diff
@@ -97,7 +97,7 @@ futures-util = "0.3.30"
 testdir = "0.9.1"
 
 [features]
-default = ["fs-store", "rpc", "net_protocol"]
```
rklaehn, line 100: todo: remove

rklaehn:
We want to expose people to the router API in the examples instead of having a self-contained Node that is not extendable.

PR #17, opened 2024-11-25
feat: Pluggable gc with exemptions

rklaehn:
## Description

This adds gc to blobs so it does not have to be driven externally. A big complication is that we want to have the ability to add exemptions to gc, e.g. from docs or willow.

So the gc is initially not running when the Blobs object is created. We have the ability to add exemption callbacks in this stage. Once the setup is done you can (but don't have to!) start the gc loop, at which point adding exemptions is no longer possible.

# Todo:

- [x] try out with docs

## Change checklist

- [x] Self-review.
- [x] Tests if relevant.

### DIFF
--- a/tests/blobs.rs
+++ b/tests/blobs.rs
@@ -59,8 +59,8 @@ async fn blobs_gc_protected() -> TestResult<()> {
     tokio::time::sleep(Duration::from_millis(100)).await;
     // protected from gc due to tag
     assert!(client.has(h1.hash).await?);
-    client.tags().delete(h1.tag).await?;
     protected.lock().unwrap().push(h1.hash);
+    client.tags().delete(h1.tag).await?;
     tokio::time::sleep(Duration::from_millis(100)).await;
     // protected from gc due to being in protected set
     assert!(client.has(h1.hash).await?);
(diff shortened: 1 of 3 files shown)


rklaehn:
The place where I tried this out is https://github.com/n0-computer/iroh-docs/tree/try-out-gc

dignifiedquire:
new api looks good

tests/blobs.rs
```diff
@@ -0,0 +1,70 @@
...
+    let protected2 = protected.clone();
+    blobs.add_protected(Box::new(move |x| {
+        let protected = protected2.clone();
+        Box::pin(async move {
+            let protected = protected.lock().unwrap();
+            for h in protected.as_slice() {
+                x.insert(*h);
+            }
+        })
+    }))?;
```
matheus23: Super minor nit:
```suggestion    let protected: Arc<Mutex<Vec<Hash>>> = Arc::new(Mutex::new(Vec::new()));
    blobs.add_protected(Box::new({
        let protected = protected.clone();
        move |x| {
            let protected = protected.clone();
            Box::pin(async move {
                let protected = protected.lock().unwrap();
                for h in protected.as_slice() {
                    x.insert(*h);
                }
            })
        }
    }))?;
```

Otherwise this seems fine.

matheus23 (approved):
I'm *fairly* confident this API will work for willow.

PR #18, opened 2024-11-28
feat: update to latest iroh

dignifiedquire:
Updates the code use the renamed `iroh-net` and dependencies

### DIFF
--- a/src/util/fs.rs
+++ b/src/util/fs.rs
@@ -126,20 +126,20 @@ pub fn relative_canonicalized_path_to_string(path: impl AsRef<Path>) -> anyhow::
     canonicalized_path_to_string(path, true)
 }
 
-/// Loads a [`iroh_net::key::SecretKey`] from the provided file, or stores a newly generated one
+/// Loads a [`iroh::key::SecretKey`] from the provided file, or stores a newly generated one
 /// at the given location.
 #[cfg(feature = "rpc")]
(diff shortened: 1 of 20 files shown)


src/util/fs.rs
```diff
@@ -126,20 +126,20 @@ pub fn relative_canonicalized_path_to_string(path: impl AsRef<Path>) -> anyhow::
     canonicalized_path_to_string(path, true)
 }
 
-/// Loads a [`iroh_net::key::SecretKey`] from the provided file, or stores a newly generated one
+/// Loads a [`iroh::key::SecretKey`] from the provided file, or stores a newly generated one
 /// at the given location.
 #[cfg(feature = "rpc")]
 #[cfg_attr(iroh_docsrs, doc(cfg(feature = "rpc")))]
-pub async fn load_secret_key(key_path: PathBuf) -> anyhow::Result<iroh_net::key::SecretKey> {
+pub async fn load_secret_key(key_path: PathBuf) -> anyhow::Result<iroh::key::SecretKey> {
```
matheus23, line 133: Don't we have this in iroh-node-utils now?

matheus23 (approved):

src/util/fs.rs
dignifiedquire: yes, we do, but it is not a dependency of iroh-blobs yet

PR #21, opened 2024-11-28
docs: Add simple file transfer example

matheus23:
## Description

Adds a simple example that basically rebuilds sendme, but simpler. No fancy clap interface, no progress bars, all in-memory blobs, copying stuff around needlessly.
But most importantly: It's very simple, just one function.

---

Also: I noticed there's an `example-iroh` feature that's enabled-by-default in blobs, which pulls in `clap`, `indicativ` and `console`. We probably don't want that in there, right?

---

I guess it would be nice to add some context about **"Why am I doing this?"**: I'm thinking of linking to this from the quickstart guide of a new iroh docs page.

## Breaking Changes

None

## Notes & open questions

Only that I removed the `example-iroh` feature from the default feature set.

## Change checklist

- [x] Self-review.
- [x] Tests if relevant.

### DIFF
--- a/examples/transfer.rs
+++ b/examples/transfer.rs
@@ -6,7 +6,6 @@ use iroh_base::ticket::BlobTicket;
 use iroh_blobs::{
     net_protocol::Blobs,
     rpc::client::blobs::{ReadAtLen, WrapOption},
-    store::mem,
     util::{local_pool::LocalPool, SetTagOption},
 };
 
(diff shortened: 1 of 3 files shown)


examples/transfer.rs
```diff
@@ -0,0 +1,100 @@
...
+            println!("File analyzed. Fetch this file by running:");
+            println!("cargo run --example transfer -- receive {ticket} {path}");
+
+            tokio::signal::ctrl_c().await?;
+        }
+        [_cmd, "receive", ticket, path] => {
+            let path_buf = PathBuf::from_str(path)?;
+            let ticket = BlobTicket::from_str(ticket)?;
+
+            let blobs = node
```
rklaehn: I guess you want to show how you can get blobs from the router, but if this was production code I would just keep the blobs from above around.

As I mentioned in https://github.com/n0-computer/iroh/pull/2938 I am not convinced that the downcast stuff is pulling its weight.

rklaehn (approved):

examples/transfer.rs
matheus23: You linked to the wrong PR, but yeah, [I agree](https://github.com/n0-computer/iroh/pull/2966#issuecomment-2507406074).

matheus23:
Seems like there was some flakyness. I filed https://github.com/n0-computer/iroh-blobs/issues/22

PR #27, opened 2024-12-04
fix!: make `net_protocol` feature work without `rpc`

dignifiedquire:
also removes `rpc` from the default feature flags to reduce dependency load

### DIFF
--- a/src/net_protocol.rs
+++ b/src/net_protocol.rs
@@ -3,12 +3,7 @@
 // TODO: reduce API surface and add documentation
 #![allow(missing_docs)]
 
-use std::{
-    collections::{BTreeMap, BTreeSet},
-    fmt::Debug,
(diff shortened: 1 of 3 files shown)


src/net_protocol.rs
```diff
@@ -187,10 +183,11 @@ impl<S: crate::store::Store> Blobs<S> {
             events,
             downloader,
             endpoint,
+            #[cfg(feature = "rpc")]
             batches: Default::default(),
             gc_state: Default::default(),
             #[cfg(feature = "rpc")]
-            rpc_handler: Arc::new(OnceLock::new()),
+            rpc_handler: Arc::new(Default::default()),
```
matheus23:
```suggestion
            rpc_handler: Default::default(),
```

rklaehn (approved):

PR #30, opened 2024-12-06
feat!: Make blobs more cheaply cloneable by by giving it an Inner

rklaehn:
## Description

Make blobs more cheaply cloneable by by giving it an Inner

Instead of having lots of Arcs in Blobs, introduce a BlobsInner and arc the whole thing. This makes cloning Blobs much cheaper and also simplifies the code a bit.

Also remove the lazily initialized rpc client/rpc handler pair and instead provide a fn to return that handler for the API user to store somewhere.

## Notes & open questions

How should the fn be called that returns the rpc handler. Probably best to rename it since just changing .client() to return an owned thing instead of a borrowed thing would result in people using it in a wrong way (create it on demand on each call, for example).

`create_client()`?
`create_handler()`?

Likewise we now need a name for the thing that is returned. People want to store it somewhere, so it can't just be an impl Deref<MemClient>.

## Change checklist

- [ ] Self-review.
- [ ] Tests if relevant.

### DIFF
--- a/src/rpc.rs
+++ b/src/rpc.rs
@@ -2,12 +2,13 @@
 
 use std::{
     io,
+    ops::Deref,
     sync::{Arc, Mutex},
 };
 
 use anyhow::anyhow;
 use client::{
-    blobs::{self, BlobInfo, BlobStatus, IncompleteBlobInfo, WrapOption},
+    blobs::{BlobInfo, BlobStatus, IncompleteBlobInfo, MemClient, WrapOption},
     tags::TagInfo,
     MemConnector,
 };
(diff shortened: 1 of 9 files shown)


src/rpc.rs
```diff
@@ -62,13 +63,8 @@ const RPC_BLOB_GET_CHANNEL_CAP: usize = 2;
...
 impl<D: crate::store::Store> Blobs<D> {
     /// Get a client for the blobs protocol
-    pub fn client(&self) -> blobs::MemClient {
-        let client = self
-            .rpc_handler
-            .get_or_init(|| RpcHandler::new(self))
-            .client
-            .clone();
-        blobs::Client::new(client)
+    pub fn client(&self) -> RpcHandler {
```
dignifiedquire: should we add a `[must_use]` here, maybe?

PR #36, opened 2024-12-10
fix: fix silent failure to add data of more than ~16MB via add_bytes or add_bytes_named

rklaehn:
## Description

- Make sure an error in send_all is not silently ignored but at least communicated to the remote side.
- Split large Bytes in add_bytes into chunks so we don't exceed MAX_FRAME_SIZE of 1024 * 1024 * 16
- Add some tests that use quinn

## Breaking Changes

None

## Change checklist

- [ ] Self-review.
- [ ] Tests if relevant.


Cargo.lock
```diff
@@ -363,6 +372,15 @@ version = "0.1.2"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "597bb81c80a54b6a4381b23faba8d7774b144c94cbd1d6fe3f1329bd776554ab"
 
+[[package]]
+name = "bincode"
```
dignifiedquire, line 376: why is this making a comback?

rklaehn: what? How did this get here?

rklaehn: Ok, it was via quic-rpc, where I forgot to remove the feature from tokio-serde.

It is gone now:
```
❯ cargo tree --all-features -i bincode
error: package ID specification `bincode` did not match any packages
```

matheus23: But the `Cargo.lock` file here still has `bincode` here, right?
Why does `cargo tree` tell you it's not in there, but the `Cargo.lock` file in this PR still adds it?

dignifiedquire: @rklaehn did you give up fixing this?

rklaehn: will be in the next PR.

PR #38, opened 2024-12-12
feat: import iroh_base::ticket::BlobTicket and iroh_base::hash

dignifiedquire:
## Description

As `iroh_base::hash` and `iroh_base::ticket::BlobtTicket` are mainly `iroh-blobs` concerns they are moved into here.

This also updates to the latest `iroh@main` so 
- `AddrInfoOptions` has been removed
- MSRV is bumped to `1.81`

## Breaking Changes

- `addr_options` on the cli is removed from the `share` call

## Change checklist

- [ ] Self-review.
- [ ] Tests if relevant.

### DIFF
--- a/src/hash.rs
+++ b/src/hash.rs
@@ -1,0 +1,564 @@
+//! The blake3 hash used in Iroh.
+
+use std::{borrow::Borrow, fmt, str::FromStr};
+
+use iroh_base::base32::{self, parse_array_hex_or_base32, HexOrBase32ParseError};
+use postcard::experimental::max_size::MaxSize;
+use serde::{de, Deserialize, Deserializer, Serialize, Serializer};
+
+/// Hash type used throughout.
+#[derive(PartialEq, Eq, Copy, Clone, Hash)]
+pub struct Hash(blake3::Hash);
+
+impl fmt::Debug for Hash {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        f.debug_tuple("Hash").field(&DD(self.to_hex())).finish()
+    }
+}
+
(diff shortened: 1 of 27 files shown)


dignifiedquire:
MSRV fixes are coming with https://github.com/n0-computer/iroh/pull/3033 and follow ups

dignifiedquire:
Ref https://github.com/n0-computer/iroh/pull/3036

src/hash.rs
```diff
@@ -0,0 +1,564 @@
...
+        self.0.as_bytes().cmp(other.0.as_bytes())
+    }
+}
+
+impl fmt::Display for Hash {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        // result will be 52 bytes
+        let mut res = [b'b'; 52];
+        // write the encoded bytes
+        data_encoding::BASE32_NOPAD.encode_mut(self.as_bytes(), &mut res);
```
rklaehn, line 127: Not for this PR, but didn't we decide to get rid of base32 here? In that case we could probably just reexport the blake3 Hash type.

src/hash.rs
```diff
@@ -0,0 +1,564 @@
...
+}
+
+#[cfg(feature = "redb")]
+mod redb_support {
+    use postcard::experimental::max_size::MaxSize;
+    use redb::{Key as RedbKey, Value as RedbValue};
+
+    use super::{Hash, HashAndFormat};
+
+    impl RedbValue for Hash {
```
rklaehn, line 238: Ah, right. This is one reason to not just reexport blake3::Hash. But we could wrap Hash in a newtype just for the store.

PR #41, opened 2024-12-16
feat!: update to iroh@0.30.0

dignifiedquire:
TODO

- [x] update to released version

### DIFF
--- a/examples/hello-world-fetch.rs
+++ b/examples/hello-world-fetch.rs
@@ -56,8 +56,7 @@ async fn main() -> Result<()> {
         "node relay server url: {:?}",
         node.endpoint()
             .home_relay()
-            .initialized()
-            .await
+            .get()?
             .expect("a default relay url should be provided")
             .to_string()
     );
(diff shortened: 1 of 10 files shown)


examples/hello-world-fetch.rs
```diff
@@ -56,6 +56,8 @@ async fn main() -> Result<()> {
         "node relay server url: {:?}",
         node.endpoint()
             .home_relay()
+            .initialized()
+            .await
             .expect("a default relay url should be provided")
             .to_string()
```
matheus23:
```suggestion
        node.endpoint()
            .home_relay()
            .get()?
            .expect("a default relay url should be provided")
            .to_string()
```

matheus23: Or, alternatively only use `node.endpoint().home_relay().initialized().await?.to_string()` without the `expect`.

dignifiedquire: fixed

ramfox (approved):

PR #42, opened 2025-01-02
refactor: Consolidate the blobs API

rklaehn:
## Description

The API has grown quite a bit. This is an attempt to structure it a bit better.

There is going to be a much larger change later when we add multi-provider downloads. That requires changing the entire progress story. But for now this just attempts a small clean up.

## Breaking Changes

- fns that import into or export from the db are moved into the store module
  - get::db::blob_info moved to store
  - get::db::valid_ranges moved to store
  - get::db::get_to_db moved to store
  - get::db::get_to_db_in_steps moved to store
  - get::db::GetStateNeedsConn moved to store
  - get::db::BlobId moved to store
  - get::db::BlobInfo moved to store
  - get::db::DownloadProgress moved to store
  - get::db::GetState moved to store
  - export::export_blob moved to store
  - export::export moved to store
  - export::export_collection moved to store
  - export::ExportProgress moved to store
- format/collection is feature gated
- DownloadProgress renamed to DownloadProgressEvent and exported in rpc/client/blobs
- BlobProgress renamed to BlobProgressEvent and exported in rpc/client/blobs

## Notes & open questions

Maybe not for this PR, but there is one open question. Basically to OO or not to OO.

There are a large number of fns that need a store. Currently these are fns that just take a impl Store as first or second parameter. I could make these member functions of the trait. E.g. handle_connection(store, conn, events) could be store.handle_connection(conn, events). Would that be easier to discover for people?

## Change checklist

- [x] Self-review.
- [ ] Tests if relevant.


src/downloader/get.rs
```diff
@@ -7,20 +7,20 @@ use iroh::endpoint;
 
 use super::{progress::BroadcastProgressSender, DownloadKind, FailureAction, GetStartFut, Getter};
 use crate::{
-    get::{db::get_to_db_in_steps, error::GetError},
-    store::Store,
+    get::Error,
+    store::{get_to_db_in_steps, FetchState, FetchStateNeedsConn, Store},
```
dignifiedquire, line 11: so there is `get`  and `fetch`  now? this is confusing 😅

src/get/progress.rs
```diff
@@ -183,3 +183,90 @@ impl TransferState {
         }
     }
 }
+
+/// Progress updates for the get operation.
+#[derive(Debug, Clone, Serialize, Deserialize)]
+pub enum DownloadProgress {
```
dignifiedquire: can we normalize `get`, `download` and `fetch`? they sound like the same thing for most users

rklaehn: Normalize to what? get, download or fetch? I guess the most descriptive is download, given that it is a downloader. 

But then get is more accurate for the case where the stuff is already local.

src/downloader/get.rs
rklaehn: everything that needs a store is in the store module. I might even make it a fn on the store traits (see the OO or not OO question)

One reason for this is that I think it would be neat if you could hide the entire store behind a ff, but to do that you must make sure that store stuff is not all over the place, otherwise you will go crazy.

dignifiedquire:
@rklaehn needs a rebase

dignifiedquire:
> There are a large number of fns that need a store. Currently these are fns that just take a impl Store as first or second parameter. I could make these member functions of the trait. E.g. handle_connection(store, conn, events) could be store.handle_connection(conn, events). Would that be easier to discover for people?

I think so. I would suggest moving everything that is expected to be used commonly on the trait, or a `StoreExtTrait`

src/get/progress.rs
dignifiedquire: I think `fetch` would work reasonably well (given that is how the browser calls this too)  and then adding suffixes if you talk about expliclit local or remote fetching

- `fetch` - could be remote or local
- `fetch_local` only reads local data
- `fetch_remote` only reads data from other peers

rklaehn:
I am not doing this. I am working on a more thorough rework of the store API that deemphasizes the rpc stuff and tries to get rid of the trait hierarchy. The trait hierarchy is too complex. It makes implementing stores that don't have the full functionality simpler, but forces quite some complexity on the end users (the store type parameter to avoid boxing, the local pool).

rklaehn:
Keeping the branch just in case...

PR #47, opened 2025-01-17
feat!: simplify LocalPool handling

dignifiedquire:
## Description

It has been a footgun for a few users, including myself, to make sure to keep around the `LocalPool`. This changes the behaviour to construct a `LocalPool` and keep it around by default. If necessary, in the builder one can provide a custom handle, if there is the need for a custom pool.

## Breaking Changes

- remove `net_protocol::Blobs::new`, use the builder instead
- remove the `LocalPoolHandle` argument from `net_protocol::Builder::build`

## Change checklist

- [ ] Self-review.
- [ ] Tests if relevant.

### DIFF
--- a/src/net_protocol.rs
+++ b/src/net_protocol.rs
@@ -241,7 +241,7 @@ impl<S: crate::store::Store> Blobs<S> {
     }
 
     pub fn rt(&self) -> &LocalPoolHandle {
-        &self.inner.rt
+        self.inner.rt()
     }
 
     pub fn downloader(&self) -> &Downloader {
--- a/src/net_protocol.rs
+++ b/src/net_protocol.rs
@@ -3,7 +3,12 @@
 // TODO: reduce API surface and add documentation
 #![allow(missing_docs)]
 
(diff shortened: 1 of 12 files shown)


src/net_protocol.rs
```diff
@@ -128,13 +157,23 @@ impl<S: crate::store::Store> Builder<S> {
...
+
     /// Build the Blobs protocol handler.
-    /// You need to provide a local pool handle and an endpoint.
-    pub fn build(self, rt: &LocalPoolHandle, endpoint: &Endpoint) -> Blobs<S> {
+    /// You need to provide a the endpoint.
+    pub fn build(self, endpoint: &Endpoint) -> Blobs<S> {
+        let rt = self
+            .rt
+            .map(Rt::Handle)
+            .unwrap_or_else(|| Rt::Owned(LocalPool::default()));
```
matheus23, line 172:
```suggestion
            .map_or_else(|| Rt::Owned(LocalPool::default()), Rt::Handle);
```

But like... could also be a `match` statement.

matheus23 (approved):
IMHO this is an obvious improvement, so I'd love to see it land.

PR #53, opened 2025-02-05
example: simplify transfer example

dignifiedquire:
## Change checklist

- [ ] Self-review.
- [ ] Tests if relevant.

### DIFF
--- a/src/hash.rs
+++ b/src/hash.rs
@@ -5,6 +5,8 @@ use std::{borrow::Borrow, fmt, str::FromStr};
 use postcard::experimental::max_size::MaxSize;
 use serde::{de, Deserialize, Deserializer, Serialize, Serializer};
 
+use crate::store::ExportFormat;
+
 /// Hash type used throughout.
 #[derive(PartialEq, Eq, Copy, Clone, Hash)]
 pub struct Hash(blake3::Hash);
@@ -239,6 +241,15 @@ impl BlobFormat {
     /// Is hash seq format
     pub const fn is_hash_seq(&self) -> bool {
         matches!(self, BlobFormat::HashSeq)
+    }
+}
+
+impl From<BlobFormat> for ExportFormat {
+    fn from(value: BlobFormat) -> Self {
+        match value {
+            BlobFormat::Raw => ExportFormat::Blob,
+            BlobFormat::HashSeq => ExportFormat::Collection,
+        }
     }
 }
 
--- a/examples/transfer.rs
+++ b/examples/transfer.rs
@@ -1,10 +1,11 @@
-use std::{path::PathBuf, str::FromStr};
+use std::path::PathBuf;
 
-use anyhow::Result;
+use anyhow::{bail, Result};
 use iroh::{protocol::Router, Endpoint};
 use iroh_blobs::{
     net_protocol::Blobs,
-    rpc::client::blobs::{ReadAtLen, WrapOption},
+    rpc::client::blobs::{self, WrapOption},
+    store::ExportMode,
     ticket::BlobTicket,
     util::SetTagOption,
 };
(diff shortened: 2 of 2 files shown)


matheus23 (commented):
I like the `node` -> `router` rename, extracting the `args` into a variable with a comment in advance (and skipping the cmd), but I'm not convinced that parsing the args twice (once in `main`, checking the length and matching on `args[0]`, then in e.g. `receive` again, checking the length again).
I think having a match like `["send", path] => ...` or `["receive", ticket, path] => ...` is much easier to read.
I'm also not super convinced that pulling out `send` and `receive` into functions is easier to understand. These are functions that are only called once, and in addition, they even have a hidden contract with the caller: E.g. in `send` it is assumed that `args[1]` won't panic, i.e. the `args` length had been checked before. Perhaps it would be better to separate the argument parsing in `main` from `send` and `receive` and pass the parsed arguments into these functions, if we want to preserve the split out functions.

src/hash.rs
```diff
@@ -242,6 +244,15 @@ impl BlobFormat {
     }
 }
 
+impl From<BlobFormat> for ExportFormat {
+    fn from(value: BlobFormat) -> Self {
+        match value {
+            BlobFormat::Raw => ExportFormat::Blob,
+            BlobFormat::HashSeq => ExportFormat::Collection,
```
dignifiedquire: @rklaehn this is pretty confusing, at least to me, can we simplify/normalize this somehow?

dignifiedquire:
> but I'm not convinced that parsing the args twice (once in main, checking the length and matching on args[0], then in e.g. receive again, checking the length again)

well it's not reparsing, just checking in stages, given the information available

dignifiedquire:
> I'm also not super convinced that pulling out send and receive into functions is easier to understand. 

that's fair

matheus23:
> > I think having a match like ["send", path] => ... or ["receive", ticket, path] => ... is much easier to read.
> 
> I have never seen anyone use this to parse arguments tbh, and was very confused when I read it, which is why I changed it

Well, I guess the haskeller in me came out. But like, this is just pattern matching, right? What specifically was confusing? Isn't this a really nice way to match CLI args, it looks a lot like the writing down... well.. the pattern.

I mean, if others agree - fine, but IMO the pattern matching reads better.

matheus23:
> > But like, this is just pattern matching, right? What specifically was confusing?
> 
> the line before for the most part
> 
> ```
> let args = std::env::args().collect::<Vec<_>>();
>     match &args.iter().map(String::as_str).collect::<Vec<_>>()[..] {
> ```
> 
> is really hard to parse and understand

Ah okay, perfect. Can we improve that part and keep the pattern matching? :pleading_face:

matheus23 (commented):
@dignifiedquire would you agree with this set of changes?

This now
- renames `node` to `router`
- extracts out `args_refs` and explains it with a comment (and uses `.as_slice()` instead of `&[..]`)
- skips the `_cmd` in the args
- generally uses `let x: X = str.parse()` instead of `X::from_str`
- uses `std::path::absolute` instead of `.canonicalize` (I seem to get errors with canoncialize?)
- uses `blobs.client().export` instead of `tokio::io::copy`ing the file out

---

I decided against extracting out the functions, and also opted for the pattern matching.

And I went for `.finish().await?` instead of just `.await?` on the progress structs. That one I'm not super sure on. I do think being explicit here is helpful. If people have this open in an editor, they can click on the `.finish()` and see where it comes from - or they can search on docs.rs for structs that have such a method and will see results.

Thoughts?

src/hash.rs
rklaehn: Yes, but not in this PR. I will keep it in mind for the blobs v1 api. But in general I don't want collections to be a central part of the store, so some mapping will still have to be done.

matheus23: FWIW, the current version of the example that I pushed (I took over this branch) doesn't depend on this additional change.
So essentially I can remove the `From` impl. I'll push a commit.

dignifiedquire:
LGTM 🚢

dignifiedquire:
> And I went for `.finish().await?` instead of just `.await?` on the progress structs.

they both suck, I am fine either way, we will make this API better in the future (TM)

matheus23 (approved):
Converting dig's approval into something that github accepts :upside_down_face:

PR #58, opened 2025-02-15
chore: patch to use main branch of iroh dependencies

ramfox:
This PR updates the following dependencies to use their main branches:

- `iroh` from `https://github.com/n0-computer/iroh.git`
- `quic-rpc` from `https://github.com/n0-computer/quic-rpc.git`
- `iroh-base` from `https://github.com/n0-computer/iroh.git`

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -51,7 +51,7 @@ dependencies = [
  "cfg-if",
  "once_cell",
  "version_check",
- "zerocopy",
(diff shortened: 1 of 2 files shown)


rklaehn:
This is meant to go together with a CI job that will run `cargo update && cargo check` to alert us if there is a breaking change, so we don't have to rush to fix things on release day.

rklaehn (approved):

PR #62, opened 2025-02-21
example: Multiprovider downloader

rklaehn:
## Description

Add the current state of the multiprovder downloader as part of an example.

The multiprovider downloader requires infrastructure from the store that as of now does not exist, specifically the ability to watch bitfields for blobs. I am working on a store refactor that will add this, but to try out the multiprovider downloader I just basically faked it for now.

For usage see the example README.md

## Breaking Changes

None

## Change checklist

- [ ] Self-review.
- [ ] Tests if relevant.


src/get/db.rs
```diff
@@ -252,7 +252,7 @@ pub async fn valid_ranges<D: MapMut>(entry: &D::EntryMut) -> anyhow::Result<Chun
     }
     let valid: ChunkRanges = valid_from_data.intersection(&valid_from_outboard);
     log!("valid_from_data: {:?}", valid_from_data);
-    log!("valid_from_outboard: {:?}", valid_from_data);
+    log!("valid_from_outboard: {:?}", valid_from_outboard);
```
rklaehn, line 255: Just improves the logging, fixes a bug in logging.

examples/multiprovider/downloader2/planners.rs
```diff
@@ -0,0 +1,233 @@
...
+/// A download planner has the option to be stateful and keep track of plans
+/// depending on the hash, but many planners will be stateless.
+///
+/// Planners can do whatever they want with the chunk ranges. Usually, they
+/// want to deduplicate the ranges, but they could also do other things, like
+/// eliminate gaps or even extend ranges. The only thing they should not do is
+/// to add new peers to the list of options.
+pub trait DownloadPlanner: Send + std::fmt::Debug + 'static {
+    /// Make a download plan for a hash, by reducing or eliminating the overlap of chunk ranges
+    fn plan(&mut self, hash: Hash, options: &mut BTreeMap<NodeId, ChunkRanges>);
```
dignifiedquire, line 19: why is this called `options`?

src/get/db.rs
```diff
@@ -241,7 +241,7 @@ pub async fn valid_ranges<D: MapMut>(entry: &D::EntryMut) -> anyhow::Result<Chun
     // compute the valid range from just looking at the data file
     let mut data_reader = entry.data_reader().await?;
     let data_size = data_reader.size().await?;
-    let valid_from_data = ChunkRanges::from(..ChunkNum::full_chunks(data_size));
+    let valid_from_data = ChunkRanges::from(..ChunkNum::chunks(data_size));
```
dignifiedquire, line 244: what about this change?

examples/multiprovider/downloader2/planners.rs
rklaehn: options to download from. What goes in is the set of options to download from, once plan is done it contains the actual plan. But agree, it is not a great name!

src/get/db.rs
rklaehn: There was a bug that led to valid_from_data being sometimes 1 chunk shorter than needed. This does not really affect you in the sendme case - it will just download 16 KiB more when resuming. But now that we want to be precise about chunk ranges, it does matter.

However, this whole valid_from_data is a hack/heuristic anyway. It makes the assumption that the data is contiguous until the size, which is true for sendme and also for normal iroh-blobs usage with the current API. But it is *definitely* no longer the case once you can have gaps in the valid range. This is one of many reasons for my current rabbit hole of redesigning the store to take bitfields into account properly including persistence.

As mentioned in the blog post, for small to medium blobs it is perfectly fine to just recompute the availabiltiy bitfield froms cratch. And it has the advantage that this is always correct, even if you do a bitfilp in the data or outboard. But for very large blobs you don't want to do it.

okdistribute:
@rklaehn do we think this is dead or will it be revived?

Issue #65, opened 2025-03-07
feat: add api to get tag info by name


dignifiedquire:
Need sth like 

```rust
let tag = blobs_client.tags().get(b"my tag").await?;
```

to get the hash belonging to a tag, and to know if such a tag exists.

PR #69, opened 2025-03-10
feat: Richer tags api

rklaehn:
## Description

This makes the anemic tags api much more rich

- Tag deletion
  - delete by range, by prefix, single item, or all 💣 
- Tag listing
  - list by range, prefix or get single item
- Tag setting
  - set a tag to a value
- Tag renaming
  - atomic rename, just like for files

## Breaking Changes

Breaks the RPC protocol and the store trait

RPC protocol:

- rpc::proto::tags::DeleteRequest now has from and to field
- rpc::proto::tags::DeleteRequest no longer has name field
- rpc::proto::tags::ListRequest now has from and to field
- rpc::proto::tags::ListRequest::all removed
- rpc::proto::tags::ListRequest::raw removed
- rpc::proto::tags::ListRequest::hash_seq removed

Store trait:

- store::traits::ReadableStore::tags now takes from and to

## Notes & open questions

Question: should we enable the rpc feature flag by default? Given that the only way to get the pleasant API at this time is the rpc, and we want beginners to have a good experience? In my big refactor you will be able to get this API without rpc enabled, but for now this is just how it is.

Implements https://github.com/n0-computer/iroh-blobs/issues/65

## Change checklist

- [x] Self-review.
- [x] Tests if relevant.

### DIFF
--- a/src/hash.rs
+++ b/src/hash.rs
@@ -256,6 +256,12 @@ pub struct HashAndFormat {
     pub format: BlobFormat,
 }
 
+impl From<Hash> for HashAndFormat {
+    fn from(hash: Hash) -> Self {
+        Self::raw(hash)
+    }
+}
+
 #[cfg(feature = "redb")]
 mod redb_support {
     use postcard::experimental::max_size::MaxSize;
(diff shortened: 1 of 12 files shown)


dignifiedquire:
would be nice to have some tests, to make sure the basics of this work

rklaehn:
> would be nice to have some tests, to make sure the basics of this work

Yeah, this is why this is a draft. But you like the API?

rklaehn:
Added some tests

PR #72, opened 2025-03-11
feat: ability to open a hash as a seekable bao file

rklaehn:
## Description

Adds the ability to open a *partial or complete* hash as a file that supports std::io::Read and std::io::Seek. If you read at a position where the data is not present, you will get an io error, since all reads are validated.

The created file will be completely independent of the store, you can use it even when the store is dropped.

This relies on some experimental code in bao-tree. I will also support it in the new store.

The underlying implementation is using positioned_io::ReadAt, and I also have a wrapper similar to tokio::io::Cursor that wraps the file so that it impls tokio:io::AsyncRead and tokio::io::AsyncSeek.

## Breaking Changes

None?

## Notes & open questions

This is very WIP

## Change checklist

- [ ] Self-review.
- [ ] Tests if relevant.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -349,8 +349,7 @@ dependencies = [
 [[package]]
 name = "bao-tree"
 version = "0.13.0"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "f1f7a89a8ee5889d2593ae422ce6e1bb03e48a0e8a16e4fa0882dfcbe7e182ef"
+source = "git+https://github.com/n0-computer/bao-tree?branch=read_and_seek#bff3195faa89636b8bdb2ff69bf7e9d532565df6"
 dependencies = [
  "bytes",
  "futures-lite",
@@ -361,6 +360,7 @@ dependencies = [
  "range-collections",
  "self_cell",
  "smallvec",
+ "tokio",
 ]
 
 [[package]]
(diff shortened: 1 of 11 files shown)


maboesanman:
Upon reading the signature 

```rust
pub async fn open(&self, hash: Hash) -> Result<impl std::io::Read + std::io::Seek> { todo!() }
```

I would personally expect this to be pending until the entire blob is downloaded, then give me readers that won't error (except under extreme cases like file permissions changing or storage devices being physically removed), but it sounds like the implementation (as described in the PR) gives you a future which resolves much more eagerly, independent of whether the blob is downloaded. This seems like it would be surprising and potentially frustrating if passing this to some code that expects read failures to be unlikely.

Maybe this api could be adjusted to something like:

```rust
pub async fn open_partially_downloaded(&self, hash: Hash) -> Result<impl std::io::Read + std::io::Seek> {
    // current implementation of the open function
}

pub async fn open(&self, hash: Hash) -> Result<impl std::io::Read + std::io::Seek> {
    let download_complete_fut = some_future(): // resolves when the download is complete
    let reader_future = open_partially_downloaded();
    let (_, reader) = join!(download_complete_fut, reader_future).await;
    Ok(reader?)
}
```

rklaehn:
Yeah, naming is TBD. This is for the YOLO case where you just want to try to read.

PR #76, opened 2025-03-14
feat!: Allow configuring the downloader when creating a blobs protocol handler

rklaehn:
## Description

Allow configuring the downloader when creating a blobs protocol handler

we don't allow passing in the entire downloader, but all config options.

## Breaking Changes

`iroh_blobs::downloader::Downloader` now takes a `config: Config ` argument instead of `concurrency_limits: ConcurrencyLimits, retry_config: RetryConfig`. `Config` has two public field `concurrency_limits` and `retry_config`, so the migration is straightforward.

Otherwise there's only additions (a new method `downloader_config` on the `Blobs` constructor).

## Change checklist

- [ ] Self-review.
- [ ] Tests if relevant.

### DIFF
--- a/src/downloader.rs
+++ b/src/downloader.rs
@@ -335,12 +335,19 @@ pub struct Config {
 }
 
 /// Handle for the download services.
-#[derive(Clone, Debug)]
+#[derive(Debug, Clone)]
 pub struct Downloader {
+    inner: Arc<Inner>,
+}
+
+#[derive(Debug)]
+struct Inner {
     /// Next id to use for a download intent.
-    next_id: Arc<AtomicU64>,
+    next_id: AtomicU64,
     /// Channel to communicate with the service.
     msg_tx: mpsc::Sender<Message>,
+    /// Configuration for the downloader.
+    config: Arc<Config>,
 }
 
 impl Downloader {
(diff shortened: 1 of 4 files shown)


src/downloader.rs
```diff
@@ -325,13 +325,29 @@ impl Future for DownloadHandle {
...
+#[derive(Debug)]
+struct Inner {
     /// Next id to use for a download intent.
-    next_id: Arc<AtomicU64>,
+    next_id: AtomicU64,
     /// Channel to communicate with the service.
     msg_tx: mpsc::Sender<Message>,
+    /// Configuration for the downloader.
+    config: Arc<Config>,
 }
```
matheus23: `Inner` is now wrapping `mpsc::Sender`, which is clonable, and `Arc<Config>`, which is also clonable.
But then it ends up inside an `Arc` in `Downloader`.
This seems redundant?
Why not
```rs
pub struct Downloader {
  next_id: Arc<AtomicU64>,
  msg_tx: mpsc::Sender<Message>,
  config: Arc<Config>,
}
```
?

ramfox (approved):
Looks good!

ramfox:
semver check indicates breaking changes, should go in the PR description!

matheus23 (approved):
:+1: besides me wondering about the `Downloader` struct wrapping.

src/downloader.rs
rklaehn: I still have the same number of arcs as before, since I am losing the arc around AtomicU64.

And having an inner means that a downloader handle is just a pointer, so it is not just cloneable but more cheaply cloneable, and does not bloat the things it is included in.

The reason the config is in an Arc is so I can share it between the handle and deep down in the actual service, even though I don't do that yet.

dignifiedquire:
@rklaehn what's the state of this PR?

Frando:
Merged main, fixed the merge commits, and filled in the *Breaking changes*  section in the PR description.

PR #86, opened 2025-04-30
Seanaye/persistence trait

seanaye:
## Description

This MR introduces a `Persistence` trait which abstracts away the persistence layer for `Completed` bao files. This means that you can BYO persistence impl to the current fs store (e.g. S3).

## Breaking Changes

The external facing breaking changes are mostly around the "asyncification" of certain trait methods. I will update this section in more detail later.

## Notes & open questions

This is still a draft MR I am just looking for feedback on the general approach before I clean things up a bit more. e.g. There are some places that can panic right now

## Change checklist

- [ ] Self-review.
- [ ] Tests if relevant.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2086,7 +2086,7 @@ dependencies = [
 
 [[package]]
 name = "iroh-blobs"
-version = "0.34.1"
+version = "0.35.1"
 dependencies = [
  "anyhow",
  "async-channel",
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "iroh-blobs"
-version = "0.34.1"
+version = "0.35.1"
 edition = "2021"
 readme = "README.md"
 description = "blob and collection transfer support for iroh"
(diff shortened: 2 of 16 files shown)


rklaehn:
Thanks for the PR. I will have to check if the storage abstraction adds performance costs for the very important local/sync case.

It might be that the best way to deal with this use case is to have a separate store implementation that allows plugging in an async storage backend, since the local store is very performance critical.

The way I usually test this is by using sendme and then syncing a directory structure that contains a lot of tiny files, like the linux kernel.

Warning: I got some important deadlines the next weeks and won't be able to look into this until after may 18.

matthiasbeyer:
I believe this cannot be trivially rebased, right?

But I think this is crucial and should not be neglected by the iroh authors! Implementing the storage backend yourself is what downstream projects want to do most likely!

matthiasbeyer:
async storage would also be nice for non-wasm contexts, because persistence is IO and doing that async'ly is (IMHO) the way to go. Especially because streaming is then only another step (streaming objects from network to storage, which would have the potential of greatly reducing memory load).

seanaye:
Hello! Sorry for the radio silence. I intend to pick this back up. My goals are async compatible storage as well as BYOS (bring your own storage) via trait. I will likely discard the current state and then force push new work to this branch.

PR #94, opened 2025-07-02
feat: add back gc protect callback

Frando:
## Description

This adds back a way to externally protect blobs from garbage collection. It works like it did in blobs1: Users can provide an (async) callback to which a `&mut HashSet<Hash>` is passed. The callback is invoked before each gc run, and all hashes added to the set will be protected from gc during this run.

Used in https://github.com/n0-computer/iroh-docs/pull/47

## Change checklist

- [ ] Self-review.
- [ ] Tests if relevant.

### DIFF
--- a/src/store/fs/gc.rs
+++ b/src/store/fs/gc.rs
@@ -130,19 +130,39 @@ fn gc_sweep<'a>(
     })
 }
 
+/// Configuration for garbage collection.
 #[derive(derive_more::Debug, Clone)]
 pub struct GcConfig {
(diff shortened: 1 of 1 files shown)


Frando:
Can you explain a bit why you don't like the idea of externally protecting blobs from gc long-term? To me it still looks appealing - it solves the iroh-docs usecase in a quite simple and straightforward method, and I can see other crates building on block having use for this as well. If you have external datastructures that maintain their own garbage collection (like docs) it seems elegant to not have to maintain tags as well.

PR #95, opened 2025-07-03
fix: getting deleted blobs should return error

Frando:
## Description

Right now there is a bug in the store: When you delete a blob by deleting its tag,  run gc, and then do `store.get_bytes(hash)`, it will not return an error but instead an emtpy byte slice (`b""`). This, of course, is very wrong: we should never return data that doesn't match the requested hash.

This PR fixes the logic to not create empty blobs on-demand when running get or export operations.

Also adds a test that ensures that, for deleted blobs, errors are returned for `get_bytes`, `export`, `export_bao`, `export_ranges` for both the `fs` and `mem` store.

Additionally, I noticed that `Blobs::delete` and `delete_with_opts` is not actually usable as a public API because there's internal protections of hashes that make it not work as intended (it is only used from gc, where protections are cleared before). Therefore I made the function non-public to not confuse users.

## Breaking Changes

* `iroh_blobs::api::blobs::Blobs::delete` and `delete_with_opts` are no longer public. These functions did not actually delete blobs, because blobs would usually be protected. The function is now only used internally during garbage collection (where protections are cleared beforehand in a non-public way). Public users can only use garbage collection for deletion of blobs.

## Change checklist

- [ ] Self-review.
- [ ] Tests if relevant.

### DIFF
--- a/src/api/blobs.rs
+++ b/src/api/blobs.rs
@@ -302,7 +302,6 @@ impl Blobs {
             ExportBaoRequest {
                 hash: hash.into(),
                 ranges: ranges.into(),
-                create_if_missing: false,
             },
             32,
         )
--- a/src/api/proto.rs
+++ b/src/api/proto.rs
@@ -222,7 +222,6 @@ pub struct ObserveRequest {
 pub struct ExportBaoRequest {
     pub hash: Hash,
     pub ranges: ChunkRanges,
-    pub create_if_missing: bool,
 }
 
 /// Export the given ranges as chunkks, without validation.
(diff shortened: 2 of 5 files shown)


rklaehn:
This is quite an embarrassing bug. Thanks for finding this first.

I don't think the `create_if_missing` parameter is useful. I removed it again and instead handled the error in the code where it computes local data for a HashSeq. WDYT?

rklaehn:
> Not too sure about the special casing of the empty hash - I'd have to dive in deeper to analyze this, but seems fine because the fs store does it too?

The empty hash is weird. Also, I don't want to remember whether I have the empty hash or not, and it costs nothing to always have it. That's the reason for the special case.

rklaehn (approved):

PR #100, opened 2025-07-09
refactor: Change concurrency approach

rklaehn:
## Description

Before, the concurrency approach for slots / handles for individual hashes was based on a top level task pool and "slots" that were managed in a map in the main actor. There were some tricky race conditions for the case where a handle would be "revived" when it was already executing its Drop fn - which does crucial things like writing the bitfield file. Also, there was actual parallelism per hash, which is probably not beneficial at all.

Now basically each hash runs effectively single threaded, meaning that we can later go from actual mutexes to more lightweight synchronisation primitives like https://crates.io/crates/atomic_refcell . Unfortunately everything must still be Send due to the fact that we run this whole thing on a multi-threaded executor 🤷 , thank you tokio. Otherwise we could just use a [RefCell](https://doc.rust-lang.org/std/cell/struct.RefCell.html).

Now the concurrency is based on a task pool that will always contain at most a single task per hash. Multiple tasks that operate on the same hash are being handled concurrently, but not in parallel, using a `FuturesUnordered`. The drop case is handled in a cleaner way - when an actor becomes idle, it "gives back" its state to the owner - the manager actor. 

If a task is being spawned while drop runs, these tasks go into the inbox and the actor gets revived immediately afterwards. The manager also has a pool of inactive actors to prevent reallocations.

All this is abstracted away by the entity_manager.

The entire entity_manager module could at some point become a separate crate, but for now I have inlined it so I don't need to do another crate.

## Change checklist

- [ ] Self-review.
- [ ] Tests if relevant.


src/store/fs.rs
```diff
@@ -525,34 +462,22 @@ impl Actor {
...
                 trace!("{cmd:?}");
                 self.db().send(cmd.into()).await.ok();
             }
+            Command::DeleteBlobs(cmd) => {
+                trace!("{cmd:?}");
+                self.db().send(cmd.into()).await.ok();
+            }
             Command::ListBlobs(cmd) => {
                 trace!("{cmd:?}");
-                let (tx, rx) = tokio::sync::oneshot::channel();
```
rklaehn, line 537: Unrelated change!

src/store/fs.rs
```diff
@@ -471,26 +428,6 @@ impl Actor {
...
-    async fn clear_dead_handles(&mut self) {
-        let mut to_remove = Vec::new();
-        for (hash, slot) in &self.handles {
-            if !slot.is_live().await {
-                to_remove.push(*hash);
-            }
-        }
-        for hash in to_remove {
-            if let Some(slot) = self.handles.remove(&hash) {
-                // do a quick check if the handle has become alive in the meantime, and reinsert it
```
rklaehn, line 483: Getting rid of this nasty race condition handling was the whole purpose of the exercise. Also, we don't need a scheduled cleanup task anymore since the entity manager will deal with it. If an actor becomes idle it will be removed from the map after shutdown.

src/metrics.rs
```diff
@@ -4,6 +4,7 @@ use iroh_metrics::{Counter, MetricsGroup};
 
 /// Enum of metrics for the module
 #[allow(missing_docs)]
+#[allow(dead_code)]
```
dignifiedquire, line 7: why is this dead code?

src/store/fs/util/entity_manager.rs
```diff
@@ -0,0 +1,1038 @@
+#![allow(dead_code)]
```
dignifiedquire, line 1: maybe remove this?

src/store/fs/util/entity_manager.rs
```diff
@@ -0,0 +1,1038 @@
...
+            .send(ShutdownAll { tx }.into())
+            .await
+            .map_err(|_| "Failed to send shutdown command")?;
+        rx.await
+            .map_err(|_| "Failed to receive shutdown confirmation")
+    }
+}
+
+#[cfg(test)]
+mod tests {
```
dignifiedquire, line 774: I wonder if using `loom` could be helpful, to gain some more confidence in the working of this cc @matheus23

src/store/fs/util/entity_manager.rs
```diff
@@ -0,0 +1,1038 @@
...
+            while let Some(res) = tasks.join_next().await {
+                if let Err(e) = res {
+                    eprintln!("Task failed during shutdown: {e:?}");
+                }
+            }
+        }
+    }
+}
+
+/// A manager for entities identified by an entity id.
```
dignifiedquire, line 667: so you basically wrote a new actor system? 😅

src/store/fs/gc.rs
```diff
@@ -299,6 +300,7 @@ mod tests {
             let outboard_path = options.outboard_path(&bh);
             let sizes_path = options.sizes_path(&bh);
             let bitfield_path = options.bitfield_path(&bh);
+            tokio::time::sleep(Duration::from_millis(100)).await; // allow for some time for the file to be written
```
dignifiedquire, line 355: this is very racy, isn't it?

src/store/fs.rs
```diff
@@ -811,7 +811,7 @@ async fn finish_import_impl(import_data: ImportEntry, ctx: HashContext) -> io::R
         }
     }
     let guard = ctx.lock().await;
-    let handle = guard.as_ref().and_then(|x| x.upgrade());
+    let handle = guard.as_ref().map(|x| x.clone());
```
dignifiedquire, line 814: could be just `.cloned()`?

src/store/fs.rs
```diff
@@ -84,15 +83,16 @@ use bao_tree::{
 };
 use bytes::Bytes;
 use delete_set::{BaoFilePart, ProtectHandle};
+use entity_manager::{EntityManagerState, SpawnArg};
```
dignifiedquire, line 86: So this is the only usage of the `entity_manager` for now?

src/metrics.rs
rklaehn: This thing was a separate crate. I inlined it so I don't have to publish and maintain yet another crate. The dead code part is the friendly API of the crate, if you don't want to embed the entity manager into your own actor.

But I want to embed the entity manager, since I don't want client -> iroh-blobs main actor -> entity manager -> entity. Save one hop.

src/store/fs/util/entity_manager.rs
rklaehn: No, that's irpc, in a way. (At least my idea how async boundaries should be structured). This is very specialized, so it's more like a task manager.

You get concurrency but not parallelism per entity, and parallelism between separate entities. An entity here is the state for a hash.

src/store/fs/gc.rs
rklaehn: Yeah. There is some delay between when an entity becomes idle and when the bitfield gets written to disk. The bitfield is ephemeral, but can be expensive to compute. You need to call shutdown before terminating the process to make sure all bitfields are persisted, but here we are looking for the file before shutting down, so there is currently no API to wait for it.

src/store/fs.rs
rklaehn: Yeah, I wrote it generically just so I don't have to mix up all the recycling logic with the actual user code in blobs. I mean, I think it would also be an useful lib, but we can take it out and publish it later.

src/store/fs/util/entity_manager.rs
rklaehn: Loom would probably be helpful, but I am not sure how I would use it here.

These are not really tests but smoke tests. My only real tests are the usage inside blobs, for now. I guess I should at least write a few normal tests. I guess the guts hanging out state thing might be useful for that...

src/store/fs/util/entity_manager.rs
rklaehn: This thing was a separate crate. I inlined it so I don't have to publish and maintain yet another crate. The dead code part is the friendly API of the crate, if you don't want to embed the entity manager into your own actor.

But I want to embed the entity manager, since I don't want client -> iroh-blobs main actor -> entity manager -> entity. Save one hop.

src/store/fs.rs
rklaehn: Yeah, but this is gone in one of the subsequent PRs anyway, so I would prefer to keep it as is...

src/metrics.rs
rklaehn: I need to rethink what metrics to collect, many of those aren't even relevant anymore.

src/store/fs/gc.rs
rklaehn: Not sure what to do about this. I could add an await_idle(hash). but that is not really a generic API, so it would have to be just for the FS store.

src/store/fs/util/entity_manager.rs
rklaehn: I have now added some real tests.

src/store/fs/util/entity_manager.rs
```diff
@@ -0,0 +1,1322 @@
...
+///
+/// The manager provides parallelism between entities, but just concurrency within a single entity.
+/// This is useful if the entity wraps an external resource such as a file that does not benefit
+/// from parallelism.
+///
+/// The entity manager internally uses a main actor and per-entity actors. Per entity actors
+/// and their inbox queues are recycled when they become idle, to save allocations.
```
matheus23, line 674: Is recycling entity actors in the hot path? Having this optimization costs us a bunch of complexity, another trait (`Reset`), and a bunch of `recycle` fns.
Perhaps the code could be slightly easier but pretty much equally as performant without recycling?

rklaehn: It probably does not matter for the fs store, since that is bound by redb for small entries and file io for large entries. But my hope is that by having the entity actors super lightweight I can justify using this thing also for the in-mem store, and there it would *definitely* matter.

Also, the recycling kind of fell out automatically. The main complexity is that you need to give back the recv side of the entity actor inbox to the main actor (or actor state) so that it can continue to send messages while the actor is shutting down, and then reawake the actor if messages have been sent during shutdown.

Once you do that, it is kind of natural to just give back the rest of the state once the shutdown is complete, and it does not add that much complexity over what you have to do anyway to deal with the shutdown race condition.

rklaehn: The alternative would be the following: the owner gets a ShuttingDown msg, stops sending messages to the actor and instead puts them in a Vec, then on ShutdownComplete creates an entirely new actor and sends all the new messages immediately, unless the Vec is empty. Would this be simpler? A bit, but not much. The 2 stage shutdown handling can not be avoided I think.

src/store/fs/util/entity_manager.rs
dignifiedquire: right

src/store/fs/gc.rs
rklaehn: So one way to fix this is to introduce a wait_idle cb for the store, which basically waits until all entities (hashes) are idle and there are also no longer any ongoing imports.

I guess that is acceptable. In the worst case I can also do something like wait_idle(hash), but I would prefer to avoid that.

rklaehn: See https://github.com/n0-computer/iroh-blobs/pull/104/commits/dcbd83b3009c67e08f80149ac175e5ca43100b75

src/store/fs/util/entity_manager.rs
matheus23: When I remove the `allow(dead_code)`, I don't actually get any errors for dead code right now.
Is it maybe worth reconsidering this?

src/store/fs/util/entity_manager.rs
```diff
@@ -0,0 +1,1322 @@
...
+    ///
+    /// Frequently this is an `Arc<Mutex<T>>` or similar. Note that per entity
+    /// access is concurrent but not parallel, so you can use a more efficient
+    /// synchronization primitive like [`AtomicRefCell`](https://crates.io/crates/atomic_refcell) if you want to.
+    type EntityState: Default + Debug + Reset + Clone + Send + Sync + 'static;
+    /// Function being called when an entity actor is shutting down.
+    fn on_shutdown(
+        state: entity_actor::State<Self>,
+        cause: ShutdownCause,
+    ) -> impl Future<Output = ()> + Send + 'static
```
matheus23, line 56: It's probably worth noting that all these `+ Send` bounds might make it a little harder to compile to Wasm.

Just something to keep an eye on.
Also FWIW, you're using `n0_future::future::Boxed`, but which will always assume `Send`, but there's also `n0_future::boxed::BoxFuture`, which is `Send` everywhere except Wasm, where it's `!Send`.

src/store/fs/util/entity_manager.rs
```diff
@@ -0,0 +1,1322 @@
...
+        ) -> Option<impl Future<Output = ()> + Send + 'static> {
+            let (entity_handle, task) = self.get_or_create(id.clone());
+            let sender = entity_handle.send();
+            if let Err(e) =
+                sender.try_send(entity_actor::Command::Spawn(Spawn { id: id.clone(), f }))
+            {
+                match e {
+                    mpsc::error::TrySendError::Full(cmd) => {
+                        let entity_actor::Command::Spawn(spawn) = cmd else {
+                            panic!()
```
matheus23, line 403: Let's at least add an error message :)

src/store/fs/util/entity_manager.rs
```diff
@@ -0,0 +1,1322 @@
...
+                            "Entity actor inbox is full, cannot send command to entity actor {:?}.",
+                            id
+                        );
+                        // we await in the select here, but I think this is fine, since the actor is busy.
+                        // maybe slowing things down a bit is helpful.
+                        (spawn.f)(SpawnArg::Busy).await;
+                    }
+                    mpsc::error::TrySendError::Closed(cmd) => {
+                        let entity_actor::Command::Spawn(spawn) = cmd else {
+                            panic!()
```
matheus23, line 415: Same here

src/store/fs.rs
```diff
@@ -708,18 +628,98 @@ impl Actor {
...
+
+    async fn spawn(
+        self,
+        manager: &mut entity_manager::EntityManagerState<EmParams>,
+        tasks: &mut JoinSet<()>,
+    ) where
+        Self: Sized,
+    {
+        let task = manager
+            .spawn_boxed(
```
matheus23, line 655: This could use `.spawn` instead, right?

src/store/fs/util/entity_manager.rs
```diff
@@ -0,0 +1,1322 @@
...
+
+/// Argument for the `EntityManager::spawn` function.
+pub enum SpawnArg<P: Params> {
+    /// The entity is active, and we were able to spawn a task.
+    Active(ActiveEntityState<P>),
+    /// The entity is busy and cannot spawn a new task.
+    Busy,
+    /// The entity is dead.
+    Dead,
+}
```
matheus23, line 77: Two somewhat related comments:
1. Given we always ignore `Busy` and `Dead` in our implementation right now and both result in `HashSpecificCommand::on_error`, perhaps we can remove `SpawnArg` altogether and just return an error from `.spawn`/`.spawn_blocking` in the main actor? Then these functions wouldn't have to be `async` anymore, too.
2. `Busy` is misleading I think. An entity actor can be "busy" in the sense that it does work, but is not yet at 100% capacity. Perhaps another name for this could be `Overloaded`? Or `AtCapacity`?

src/store/fs/util/entity_manager.rs
```diff
@@ -0,0 +1,1322 @@
...
+                        // re-assemble the actor from the parts
+                        let mut actor = entity_actor::Actor {
+                            main: self.internal_send.clone(),
+                            recv,
+                            state,
+                            tasks,
+                        };
+                        if actor.recv.is_empty() {
+                            // No commands during shutdown, we can recycle the actor.
+                            self.recycle(send, actor);
```
matheus23, line 479: Is there a race condition here?
I struggle to understand the whole web of senders, receivers and spawns and whether there's potential for true parallelism or just concurrency... But in theory `recv` might receive an item after `is_empty()` returns true and before `self.recycle`, right?

What's the worst that could happen in that case?

Otherwise if it's impossible, perhaps it's also impossible for `actor.recv.is_empty()` to return false, in that case maybe there's potential for removing the `recv`. Not sure though.

src/store/fs/util/entity_manager.rs
matheus23: I guess it works because there's another `#![allow(dead_code)]` in the `main_actor` module.

matheus23 (approved):
LGTM. I have some nits/ideas, but maybe they just go into follow-ups.

src/store/fs/util/entity_manager.rs
matheus23: Ignore me. rust-analyzer doesn't mark them as unused, because they're used in tests.

src/store/fs/util/entity_manager.rs
rklaehn: I guess it should be unreachable!, since we are calling this with Spawn a few lines further up.

src/store/fs/util/entity_manager.rs
rklaehn: making this unreachable as well

src/store/fs.rs
rklaehn: I think I got this fixed in a subsequent PR, so leaving it alone for now.

src/store/fs/util/entity_manager.rs
rklaehn: Let's fix WASM once we get there. Or is there a SendUnlessWasm or something that I could just use here?

src/store/fs/util/entity_manager.rs
rklaehn: I like having the FnOnce being called exactly once and not at most once. Sadly we are already in callback land, so might as well embrace it. It is a matter of taste I guess.

src/store/fs/util/entity_manager.rs
rklaehn: I think it is safe. Tick is the driver for the main actor / actor state. It is called sequentially (guaranteed by `&mut self` ) and it is also guaranteed not to be called concurrently with spawn_boxed, which also takes `&mut self`.

The only place where send is called is in spawn_boxed (entity_handle.send().try_send(...)).

> Otherwise if it's impossible, perhaps it's also impossible for actor.recv.is_empty() to return false, in that case maybe there's potential for removing the recv. Not sure though.

No it is definitely possible for recv.is_empty to return true or false. E.g. you got an async shutdown fn that takes a very long time, and while this is ongoing we try to do another thing with this entity. Then the queue is non-empty (can even go all the way to full) and we have to "revive" the entity instead of recycling.

src/store/fs/util/entity_manager.rs
rklaehn: Not sure what to do about Busy. It is almost impossible to reach this. This happens if you send messages faster than they can be converted into local tasks on the recv side. I guess a burst of tasks would do it. Maybe just call it QueueFull? But then this does not really appear in the public API, so I don't lose sleep about it.

PR #101, opened 2025-07-09
refactor: Lightweight watcher

rklaehn:
## Description

In the current blobs I am using tokio::sync::watch for notifying the observers when the state for a hash has changed. This works, but is way more heavy than needed.

1. tokio::sync::watch is optimized for a case where we have tons of watchers. It uses a giant BigNotify under the hood to solve the thundering herd problem. I think it will be much more common that you have a few tasks operating on a hash (frequently just one), and a few observers (frequently none).
2. since we now know that all tasks for a hash run concurrently but not parallel, we can use the awesome AtomicRefCell instead of a full blown lock.

## Breaking Changes

None

## Change checklist

- [ ] Self-review.
- [ ] Tests if relevant.


src/store/fs.rs
```diff
@@ -217,10 +217,17 @@ impl entity_manager::Params for EmParams {
 
     type EntityState = Slot;
 
-    async fn on_shutdown(
-        _state: entity_manager::ActiveEntityState<Self>,
-        _cause: entity_manager::ShutdownCause,
-    ) {
+    async fn on_shutdown(state: HashContext, cause: ShutdownCause) {
```
rklaehn: Unrelated change!

PR #102, opened 2025-07-09
refactor: simplify state of fs store entry

rklaehn:
## Description

A few things were stored twice, just to enable Drop. Now we don't need Drop anymore and can store them only once.

BaoFileHandle was a wrapper around a BaoFileHandleInner, which had the actual BaoFileStorage as well as the additional info needed to be able to implement Drop. Now we don't implement Drop anymore and can just pass in the required extra info in the persist call from on_shutdown!

In several other places we now have to give additional info when working with BaoFileHandle, but this is conveniently grouped in HashContext.

## Breaking Changes

None

## Change checklist

- [ ] Self-review.
- [ ] Tests if relevant.

### DIFF
--- a/src/store/fs/util/entity_manager.rs
+++ b/src/store/fs/util/entity_manager.rs
@@ -400,7 +400,7 @@ mod main_actor {
                 match e {
                     mpsc::error::TrySendError::Full(cmd) => {
                         let entity_actor::Command::Spawn(spawn) = cmd else {
-                            panic!()
+                            unreachable!()
                         };
                         warn!(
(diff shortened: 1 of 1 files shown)


src/store/fs/util/entity_manager.rs
```diff
@@ -400,7 +400,7 @@ mod main_actor {
                 match e {
                     mpsc::error::TrySendError::Full(cmd) => {
                         let entity_actor::Command::Spawn(spawn) = cmd else {
-                            panic!()
+                            unreachable!()
```
rklaehn, line 403: Forgot to do this in the last PR!

PR #104, opened 2025-07-10
refactor: Remove slot mutex and simplify per blob state

rklaehn:
## Description

Before, the state we kept per hash was a bit convoluted. The entry point was a Slot, which contained a tokio mutex to cover the async loading of an entry state from the metadata db. Then inside that, there was the actual entry state, wrapped in an option to cover the case that we don't have anything about the hash.

This was working, but it was an arc in an arc in an arc, and also led to bugs in the case of trying to export a blob that doesn't exist.

Now all these states are flattened into a single enum, and we can easily define what should happen when we e.g. do an export of an entry that does not exist - return an appropriate io error.

## Breaking Changes

None

## Notes & open questions

Note: you could remove the Initial and Loading state by using a tokio::sync::OnceCell. But that is a true sync primitive, and we want to use an AtomicRefCell, and also using a OnceLock would come with its own sync primitive (a semaphore) just for init, and then we have our own one for the non-load state changes. I don't think it is that bad...

Note: a lot of changes are to add the wait_idle fn in the rpc protocol, which isn't really related to the changes in the PR but just a way to get rid of some flaky tests.

## Change checklist

- [ ] Self-review.
- [ ] Tests if relevant.


src/store/fs/util/entity_manager.rs
```diff
@@ -12,6 +12,12 @@ use tokio::sync::{mpsc, oneshot};
 pub trait Reset: Default {
     /// Reset the state to its default value.
     fn reset(&mut self);
+
+    /// A ref count to ensure that the state is unique when shutting down.
+    ///
+    /// You are not allowed to clone the state out of a task, even though that
+    /// is possible.
+    fn ref_count(&self) -> usize;
```
rklaehn: This isn't really needed - just a safeguard to alert people if they clone the state out of a task.

Not sure if I should keep it.

src/store/fs/util/entity_manager.rs
```diff
@@ -213,6 +219,8 @@ mod entity_actor {
                 )
                 .await
                 .ok();
+            println!("Calling on_shutdown {}", self.state.state.ref_count());
```
dignifiedquire: stray println

src/store/fs.rs
```diff
@@ -644,25 +687,24 @@ trait HashSpecificCommand: HashSpecific + Send + 'static {
...
+                        SpawnArg::Dead => {
+                            self.on_error(arg).await;
+                        }
+                    }
+                }
+                .instrument(span)
+            })
             .await;
         if let Some(task) = task {
             tasks.spawn(task);
```
matheus23, line 710: IMHO it would make more sense to `.instrument` here instead of above. I don't think there currently are, but in theory there could be logs between the callback to `spawn` and the future returned from `spawn` that wouldn't be instrumented properly otherwise.

src/store/fs.rs
```diff
@@ -1188,8 +1328,14 @@ impl FsStore {
 
     /// Load or create a new store with custom options, returning an additional sender for file store specific commands.
     pub async fn load_with_opts(db_path: PathBuf, options: Options) -> anyhow::Result<FsStore> {
+        static THREAD_NR: AtomicU64 = AtomicU64::new(0);
         let rt = tokio::runtime::Builder::new_multi_thread()
-            .thread_name("iroh-blob-store")
+            .thread_name_fn(|| {
+                format!(
+                    "iroh-blob-store-{}",
+                    THREAD_NR.fetch_add(1, Ordering::SeqCst)
```
matheus23:
```suggestion
                    THREAD_NR.fetch_add(1, Ordering::Relaxed)
```
No need for `SeqCst` here.

src/store/fs/util/entity_manager.rs
matheus23: This trait isn't exposed publicly.
So "people" would be us I guess.

What could happen if we accidentally kept a clone of the state? I guess it potentially gets reset via internal mutation.

I'm unsure, too.
I guess everyone who's working with this will know that entity states get recycled - that's the whole point.
So ... maybe not really needed?

src/store/fs/util/entity_manager.rs
```diff
@@ -959,10 +970,10 @@ mod tests {
                 assert_eq!(global.data, values, "Data mismatch");
                 for id in values.keys() {
                     let log = global.log.get(id).unwrap();
-                    assert!(
-                        log.len() % 2 == 0,
-                        "Log must contain alternating wakeup and shutdown events"
-                    );
+                    if log.len() % 2 != 0 {
+                        println!("{log:#?}");
```
matheus23: leftover println

src/store/fs/util/entity_manager.rs
rklaehn: This is a common problem I keep having. Things that are Clone to avoid lifetime issues, but you can't *really* freely clone them since they have internal state and can go bad.

E.g. iroh connections in a connection pool. You are not allowed to clone them out of the pool since they are supposed to be managed by the pool, but it is impossible to write an API that entirely prevents it.

src/store/fs.rs
rklaehn: I don't think that would work. The above spawn adds a "task" to the the FuturesUnordered that might already exist for that hash. Only if the actor for the task doesn't yet exist does it return a Some(...) for the caller to spawn.

src/store/fs/util/entity_manager.rs
rklaehn: Removed it for now. If it needs to exist, it does not belong in the Reset trait.

PR #105, opened 2025-07-10
docs: Update the transfer example to iroh 0.90

matheus23:
## Description

Revives the [old transfer example](https://github.com/n0-computer/iroh-blobs/blob/8a975ecda8bca5e8988db911857012e6817b456e/examples/transfer.rs) so we can use it on the https://www.iroh.computer/docs/quickstart page

## Notes & open questions

Also fixes a bug.

## Change checklist

- [x] Self-review.

### DIFF
--- a/examples/transfer.rs
+++ b/examples/transfer.rs
@@ -1,20 +1,14 @@
 use std::path::PathBuf;
 
-use anyhow::Result;
 use iroh::{protocol::Router, Endpoint};
-use iroh_blobs::{
-    api::blobs::{AddPathOptions, ExportMode, ExportOptions, ImportMode},
-    net_protocol::Blobs,
-    store::mem::MemStore,
-    ticket::BlobTicket,
-    BlobFormat,
-};
+use iroh_blobs::{net_protocol::Blobs, store::mem::MemStore, ticket::BlobTicket};
 
 #[tokio::main]
-async fn main() -> Result<()> {
+async fn main() -> anyhow::Result<()> {
     // Create an endpoint, it allows creating and accepting
     // connections in the iroh p2p world
     let endpoint = Endpoint::builder().discovery_n0().bind().await?;
+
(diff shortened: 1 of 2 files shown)


examples/transfer.rs
```diff
@@ -58,38 +39,36 @@ async fn main() -> Result<()> {
...
             let abs_path = std::path::absolute(filename)?;
             let ticket: BlobTicket = ticket.parse()?;
 
             println!("Starting download.");
 
-            downloader
+            // For receiving files, we create a "downloader" that allows us to fetch files
+            // from other nodes via iroh connections
+            store
+                .downloader(&endpoint)
```
rklaehn: For this example this is fine, but since the downloader has some internal state, it is probably best to create it once for efficiency, if people take this example as a starting point to build something more complex.

examples/transfer.rs
```diff
@@ -58,38 +39,36 @@ async fn main() -> Result<()> {
...
 
-            store
-                .blobs()
-                .export_with_opts(ExportOptions {
-                    hash: ticket.hash(),
-                    mode: ExportMode::TryReference,
-                    target: abs_path,
-                })
-                .await?;
+            store.blobs().export(ticket.hash(), abs_path).await?;
```
rklaehn, line 72: 👍 
TryReference is nice for efficiency on ancient file system, but on all modern file systems (NTFS, BTRFS, whatever OSX has) CoW will be just as efficient and less... problematic.

examples/transfer.rs
rklaehn: If you want a stateless way to download, there is also api::remote::Remote::fetch, but that requires you to create your own connection.

matheus23: Good point! I slightly changed the code to store the downloader in a variable.

rklaehn (approved):

Issue #110, opened 2025-07-11
Make sure the error kind is properly serialized in the rpc api


rklaehn:
Currently everything is "other", which is lame.

PR #111, opened 2025-07-11
feat: make sure the error kind is properly transferred when serializing io errors

rklaehn:
## Description

I now go through a lot of effort assigning good error kinds and don't want them to be lost!

## Breaking Changes

None, except the rpc API will behave slightly differently. But we don't care for that level of stability yet!

## Notes & open questions

Note: you can cast an ErrorKind to u32, but the order in the enum is not stable. Hence the boilerplatey match.

Fixes https://github.com/n0-computer/iroh-blobs/issues/110

## Change checklist

- [ ] Self-review.
- [ ] Tests if relevant.


matheus23 (approved):
Eh, I wish there were a way to check for future variants that might get added and have a test failing or sth like that. But with `#[non_exhaustive]` and no way of listing all current types of errors, I wouldn't know how.

(perhaps some hackery with `mem::transmute` between u32 and `ErrorKind` and counting up until it fails? 😅 )

rklaehn:
> Eh, I wish there were a way to check for future variants that might get added and have a test failing or sth like that. But with `#[non_exhaustive]` and no way of listing all current types of errors, I wouldn't know how.
> 
> (perhaps some hackery with `mem::transmute` between u32 and `ErrorKind` and counting up until it fails? 😅 )

Yeah, I actually tried to leave out the _ case, but the compiler shouted at me due to non_exhaustive...

PR #114, opened 2025-07-11
docs: Add module docs

rklaehn:
## Description

Adds module docs and includes the README.md into the docs so the example in README.md is being tested.

## Breaking Changes

None, I guess. The additional deref is not breaking.

## Notes & open questions

There is an unrelated change, adding back lifetimes, the lack of which that make my local clippy unhappy (I only removed the lifetimes because of clippy in the first place).

## Change checklist

- [ ] Self-review.
- [ ] Tests if relevant.

### DIFF
--- a/README.md
+++ b/README.md
@@ -22,6 +22,7 @@ This crate is used together with [iroh](https://crates.io/crates/iroh). Connecti
 
 - **Requester:** The side that asks for data. It is initiating requests to one or many providers.
 
+A node can be a provider and a requester at the same time.
 
 ## Getting started
 
@@ -31,9 +32,9 @@ Iroh provides a [`Router`](https://docs.rs/iroh/latest/iroh/protocol/struct.Rout
 
 Here is a basic example of how to set up `iroh-blobs` with `iroh`:
 
-```rust
+```rust,no_run
 use iroh::{protocol::Router, Endpoint};
-use iroh_blobs::{store::Store, net_protocol::Blobs};
+use iroh_blobs::{store::mem::MemStore, BlobsProtocol};
 
 #[tokio::main]
 async fn main() -> anyhow::Result<()> {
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -33,31 +34,31 @@ Here is a basic example of how to set up `iroh-blobs` with `iroh`:
...
     let router = Router::builder(endpoint)
         .accept(iroh_blobs::ALPN, blobs.clone())
         .spawn();
 
-    // do fun stuff with the blobs protocol!
+    let tag = blobs.add_slice(b"Hello world").await?;
+    println!("We are now serving {}", blobs.ticket(tag).await?);
+
+    // wait a second
+    tokio::time::sleep(std::time::Duration::from_secs(1)).await;
```
matheus23:
```suggestion
    // serve requests until Ctrl+C
    tokio::signal::ctrl_c().await?;
```
Perhaps less confusing?;

src/net_protocol.rs
```diff
@@ -66,6 +66,14 @@ pub struct BlobsProtocol {
     pub(crate) inner: Arc<BlobsInner>,
 }
 
+impl Deref for BlobsProtocol {
+    type Target = Store;
+
+    fn deref(&self) -> &Self::Target {
+        &self.inner.store
+    }
```
matheus23, line 74: Personally I'm not a huge fan of derefs, but also 🤷

matheus23 (approved):
Just some nits

README.md
rklaehn: Yeah, I actually had that. But this is run during tests, so tests were hanging and I was confused, it being Friday afternoon. I guess I could use no_run?

src/net_protocol.rs
rklaehn: > Personally I'm not a huge fan of derefs, but also 🤷

If we don't like derefs in public APIs, we would also have to remove the deref from MemStore and FsStore to the client API. But if we want to do this, it would have to be a separate PR since it would affect a lot of places.

PR #119, opened 2025-07-22
feat: Read and seek

rklaehn:
## Description

Add an api for reading and seeking a blob using tokio::io::AsyncRead and tokio::io::AsyncSeek

## Breaking Changes

Changes api::blobs::ExportRangesProgress::stream to be non-async - it does not have to be!

## Notes & open questions

Note: the entire machinery of seek and read is horribly undocumented regarding the non happy case. So I went with the following approach: if you ever drop a future of either seek or read before completing, the reader becomes poisioned and will never work again, you have to grab a new one, which is cheap.

This bypasses the whole question about what to do if somebody calls x without having completed y, and in normal usage you should never notice this.

## Change checklist

- [ ] Self-review.
- [ ] Tests if relevant.

### DIFF
--- a/src/api/blobs.rs
+++ b/src/api/blobs.rs
@@ -873,7 +873,7 @@ impl ExportRangesProgress {
     /// range of 0..100, you will get the entire first chunk, 0..1024.
     ///
     /// It is up to the caller to clip the ranges to the requested ranges.
-    pub async fn stream(self) -> impl Stream<Item = ExportRangesItem> {
+    pub fn stream(self) -> impl Stream<Item = ExportRangesItem> {
         Gen::new(|co| async move {
             let mut rx = match self.inner.await {
                 Ok(rx) => rx,
--- a/src/api/blobs.rs
+++ b/src/api/blobs.rs
@@ -105,10 +105,34 @@ impl Blobs {
         })
     }
 
+    /// Create a reader for the given hash. The reader implements [`tokio::io::AsyncRead`] and [`tokio::io::AsyncSeek`]
+    /// and therefore can be used to read the blob's content.
+    ///
+    /// Any access to parts of the blob that are not present will result in an error.
+    ///
+    /// Example:
+    /// ```rust
+    /// use iroh_blobs::{store::mem::MemStore, api::blobs::Blobs};
+    /// use tokio::io::AsyncReadExt;
+    ///
+    /// # async fn example() -> anyhow::Result<()> {
+    /// let store = MemStore::new();
+    /// let tag = store.add_slice(b"Hello, world!").await?;
+    /// let mut reader = store.reader(tag.hash);
+    /// let mut buf = String::new();
+    /// reader.read_to_string(&mut buf).await?;
+    /// assert_eq!(buf, "Hello, world!");
+    /// # Ok(())
+    /// }
+    /// ```
     pub fn reader(&self, hash: impl Into<Hash>) -> BlobReader {
         self.reader_with_opts(ReaderOptions { hash: hash.into() })
     }
 
+    /// Create a reader for the given options. The reader implements [`tokio::io::AsyncRead`] and [`tokio::io::AsyncSeek`]
+    /// and therefore can be used to read the blob's content.
+    ///
+    /// Any access to parts of the blob that are not present will result in an error.
     pub fn reader_with_opts(&self, options: ReaderOptions) -> BlobReader {
         BlobReader::new(self.clone(), options)
     }
--- a/src/api/blobs.rs
+++ b/src/api/blobs.rs
@@ -29,8 +29,11 @@ use n0_future::{future, stream, Stream, StreamExt};
 use quinn::SendStream;
 use range_collections::{range_set::RangeSetRange, RangeSet2};
 use ref_cast::RefCast;
+use serde::{Deserialize, Serialize};
 use tokio::io::AsyncWriteExt;
 use tracing::trace;
+mod reader;
+pub use reader::Reader;
 
 // Public reexports from the proto module.
 //
@@ -100,6 +103,14 @@ impl Blobs {
             blobs: self,
             _tx: tx,
         })
+    }
+
+    pub fn reader(&self, hash: impl Into<Hash>) -> Reader {
+        self.reader_with_opts(ReaderOptions { hash: hash.into() })
+    }
+
+    pub fn reader_with_opts(&self, options: ReaderOptions) -> Reader {
+        Reader::new(self.clone(), options)
     }
 
     /// Delete a blob.
@@ -645,6 +656,12 @@ impl<'a> AddProgress<'a> {
     pub async fn stream(self) -> impl Stream<Item = AddProgressItem> {
         self.inner
     }
+}
+
+/// Options for an async reader for blobs that supports AsyncRead and AsyncSeek.
+#[derive(Debug, Clone, Serialize, Deserialize)]
+pub struct ReaderOptions {
+    pub hash: Hash,
 }
 
 /// An observe result. Awaiting this will return the current state.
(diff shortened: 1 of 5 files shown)


src/api/blobs.rs
```diff
@@ -647,6 +658,12 @@ impl<'a> AddProgress<'a> {
     }
 }
 
+/// Options for an async reader for blobs that supports AsyncRead and AsyncSeek.
+#[derive(Debug, Clone, Serialize, Deserialize)]
+pub struct ReaderOptions {
```
rklaehn, line 687: You might think that this options object is a bit much, but who knows, maybe we want to add more options in the future.

One option would be the "magic wait until download" option, e.g. you seek to somewhere where the blob isn't downloaded yet, it waits instead of failing.

Or even, you seek to somewhere where the blob isn't downloaded yet, it **triggers a download** and waits instead of failing.

src/api/blobs.rs
```diff
@@ -102,6 +105,14 @@ impl Blobs {
         })
     }
 
+    pub fn reader(&self, hash: impl Into<Hash>) -> BlobReader {
```
dignifiedquire, line 128: would be nice to add a doctest/example for this

src/api/blobs/reader.rs
```diff
@@ -0,0 +1,328 @@
...
+            }
+        }
+        Ok(())
+    }
+
+    #[tokio::test]
+    async fn reader_partial_fs() -> TestResult<()> {
+        let testdir = tempfile::tempdir()?;
+        let store = FsStore::load(testdir.path().to_owned()).await?;
+        // reader_smoke_raw(store.blobs()).await?;
```
dignifiedquire: what's this?

rklaehn: dead code

rklaehn:
Cross test fails with "memory allocation of 8388608 bytes failed". I think this is unrelated to the changes.

src/api/blobs/reader.rs
```diff
@@ -0,0 +1,327 @@
...
+        position: u64,
+        #[debug(skip)]
+        op: n0_future::boxed::BoxStream<ExportRangesItem>,
+    },
+    #[default]
+    Poisoned,
+}
+
+impl BlobReader {
+    pub fn new(blobs: Blobs, options: ReaderOptions) -> Self {
```
dignifiedquire: should this be private?

src/api/blobs/reader.rs
```diff
@@ -0,0 +1,327 @@
...
+    Reading {
+        position: u64,
+        #[debug(skip)]
+        op: n0_future::boxed::BoxStream<ExportRangesItem>,
+    },
+    #[default]
+    Poisoned,
+}
+
+impl BlobReader {
```
dignifiedquire, line 42: it could be nice to add an option `fn hash(&self) -> Hash` or so, given that this data is available

src/api/blobs/reader.rs
```diff
@@ -0,0 +1,327 @@
...
+                        }
+                        Poll::Ready(None) => {
+                            // done with the stream, go back in idle.
+                            *guard = ReaderState::Idle {
+                                position: *position1,
+                            };
+                            break Poll::Ready(Ok(()));
+                        }
+                        Poll::Pending => {
+                            break if position != *position1 {
```
dignifiedquire, line 120: iinteresting, is this new syntax? I have never seen this

src/api/blobs/reader.rs
```diff
@@ -0,0 +1,327 @@
...
+                            break Poll::Ready(Err(io::Error::other(format!(
+                                "Error reading data: {err}"
+                            ))));
+                        }
+                        Poll::Ready(None) => {
+                            // done with the stream, go back in idle.
+                            *guard = ReaderState::Idle {
+                                position: *position1,
+                            };
+                            break Poll::Ready(Ok(()));
```
dignifiedquire, line 117: might be easier to read if you use `return `  for these? maybe that's just me

src/api/blobs/reader.rs
rklaehn: It's just a break in a loop with value, and the value is an if expression. Just my old FP style.

src/api/blobs/reader.rs
rklaehn: I had break so I could log the value before returning during debugging.

src/api/blobs/reader.rs
```diff
@@ -0,0 +1,327 @@
...
+    fn poll_read(
+        self: Pin<&mut Self>,
+        cx: &mut Context<'_>,
+        buf: &mut tokio::io::ReadBuf<'_>,
+    ) -> Poll<io::Result<()>> {
+        let this = self.get_mut();
+        let mut position1 = None;
+        loop {
+            let guard = &mut this.state;
+            match std::mem::take(guard) {
```
dignifiedquire, line 66: fyi, I had this pattern show up being extremly slow in another project :'( with the rust compiler allocating over and over again, making reads/writes 

you can see how I fixed/worked around it here: https://github.com/rpgp/rpgp/pull/579/files#diff-70038cda95dcb4bc54877b1a3eb9f111ed226310dad30909b6d3cb88af6cd2f5

not saying this is currently slow, just warning you this can be very slow

src/api/blobs/reader.rs
```diff
@@ -0,0 +1,327 @@
...
+                            "Seeking from end is not supported yet",
+                        ))?;
+                    }
+                };
+                *guard = ReaderState::Seeking {
+                    position: position1,
+                };
+                Ok(())
+            }
+            ReaderState::Reading { .. } => Err(io::Error::other("Can't seek while reading")),
```
dignifiedquire, line 186: this is a bit unfortunate, not necessary for the PR but I have had instances in the past where I wanted to say read the first 100 bytes, seek 200 bytes, and then continue reading

src/api/blobs/reader.rs
rklaehn: That is unfortunate. I like this pattern.

I think it should be fine as of now. The state is just 32 bytes, so it is very unlikely to matter compared with all the async stuff going on. I could rewrite it to just match on the state in place, but then there is always the chance that you get in some weird state instead of poisoned if you forget to handle an odd error.

src/api/blobs/reader.rs
rklaehn: You can do that, as long as you don't drop futures.

read(...).await?
seek(...).await?
read(...).await?

is fine. the only thing that would break it would be something like this:

read(...).timeout(Duration::from_millis(1)).await?; // read future is dropped.
seek(...).await;

I mean, I would love to support this, but the docs is uniquely unhelpful about what the behaviour should be.

"The position returned by calling this method can only be relied on right after start_seek. If you have changed the position by e.g. reading or writing since calling start_seek, then it is **_unspecified_** whether the returned position takes that position change into account. Similarly, if start_seek has never been called, then it is _**unspecified**_ whether poll_complete returns the actual position or some other placeholder value (such as 0)."

I guess you could make it so a seek from an absolute position (start and later end) recovers the state you get into when a read future is dropped, since at this point the internal state is deterministic again... 🤷

dignifiedquire: ah right, I missunderstood

> "The position returned by calling this method can only be relied on right after start_seek. If you have changed the position by e.g. reading or writing since calling start_seek, then it is **_unspecified_** whether the returned position takes that position change into account. Similarly, if start_seek has never been called, then it is _**unspecified**_ whether poll_complete returns the actual position or some other placeholder value (such as 0)."

uff...

src/api/blobs.rs
dignifiedquire: maybe mark it as non_exhaustive in this case?

dignifiedquire (approved):

Issue #120, opened 2025-07-23
revive examples from iroh-blobs 0.35


rklaehn:
Some of them were really iroh examples, not iroh-blobs examples. But still, there were more...

- [x] custom-protocol.rs
- [x] discovery-local-network.rs
- [ ] fetch-fsm.rs
- [x] fetch-stream.rs
- [ ] hello-world-fetch.rs
- [ ] hello-world-provide.rs
- [ ] provide-bytes.rs
- [x] transfer.rs

rklaehn:
Not sure if the hello-world-... ones provide much benefit over transfer.
Renamed the discovery-local-network one to mdns-discovery
Not sure about fetch-fsm, that is so low level...

Issue #117, opened 2025-07-16
request example is empty


dignifiedquire:
`examples/request.rs` just contains an empty `main.rs`

rklaehn:
Example was removed.

PR #121, opened 2025-07-23
docs: Revive 0.35 examples

rklaehn:
## Description

Add back the examples that were removed when moving from 0.35 to 0.9x.

## Breaking Changes

None

## Notes & open questions

How much should we bikeshed the progress reporting in the examples, using console and indicatif crates? On the one hand it is nice, on the other hand it makes the examples more complex.

I think we might put basic progress display for import, download and export into the util crate. But I think it would be nice if we could get by without the `--features exampes` feature flag for most or all examples, so maybe not use console and indicatif so we don't have to have it as dev dependencies.

I changed the transfer example to use clap like the other examples. I hope you don't mind, @matheus23.

## Change checklist

- [ ] Self-review.
- [ ] Tests if relevant.

Implements https://github.com/n0-computer/iroh-blobs/issues/120
Implements https://github.com/n0-computer/iroh-blobs/issues/117

### DIFF
--- a/examples/mdns-discovery.rs
+++ b/examples/mdns-discovery.rs
@@ -19,16 +19,9 @@ use iroh::{
     discovery::mdns::MdnsDiscovery, protocol::Router, Endpoint, PublicKey, RelayMode, SecretKey,
 };
 use iroh_blobs::{store::mem::MemStore, BlobsProtocol, Hash};
-use tracing_subscriber::{prelude::*, EnvFilter};
 
-// set the RUST_LOG env var to one of {debug,info,warn} to see logging info
-pub fn setup_logging() {
-    tracing_subscriber::registry()
-        .with(tracing_subscriber::fmt::layer().with_writer(std::io::stderr))
-        .with(EnvFilter::from_default_env())
-        .try_init()
-        .ok();
-}
+mod common;
+use common::{get_or_generate_secret_key, setup_logging};
 
 #[derive(Debug, Parser)]
 #[command(version, about)]
@@ -64,13 +57,12 @@ async fn accept(path: &Path) -> Result<()> {
     }
 
     let key = get_or_generate_secret_key()?;
-    let discovery = MdnsDiscovery::new(key.public())?;
 
     println!("Starting iroh node with mdns discovery...");
     // create a new node
     let endpoint = Endpoint::builder()
         .secret_key(key)
-        .add_discovery(discovery)
+        .add_discovery(MdnsDiscovery::builder())
         .relay_mode(RelayMode::Disabled)
         .bind()
         .await?;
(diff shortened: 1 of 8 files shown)


examples/mdns-discovery.rs
```diff
@@ -0,0 +1,144 @@
+//! Example that runs and iroh node with local node discovery and no relay server.
```
matheus23:
```suggestion
//! Example that runs an iroh node with local node discovery and no relay server.
```

examples/transfer.rs
matheus23, line 1: I appreciate trying to improve this example a bit, but any changes mean we need to update [the quickstart guide](https://www.iroh.computer/docs/quickstart) on the website, as it directly quotes the code from this example and links to it.

I originally considered doing the clap command parsing in the example, but I went without it because it's just so much more code and more concepts to explain (the Parser macro, the Subcommand macros, etc.).

Separating `send` and `receive` into their own functions is also something we've [considered in the past](https://github.com/n0-computer/iroh-blobs/pull/53#issuecomment-2642712157).

Not sure. I can see how this example doesn't read like "production rust code should look like", but it really serves another purpose by trying to have as few things as possible that need to be explained besides the iroh & iroh-blobs API itself.

Anyhow, maybe I'm wrong with some of these decisions, but in any case: I'd only accept these changes in conjunction with an update to the quickstart guide.

rklaehn: I am not really sure about the clap thing either. The code before was fine, I just don't want to do this in all the other examples.

The reason I split the send and receive path is because I dislike the (ephemeral) receive side publishing it's addr info to discovery. I think it is good to encourage people to use passive discovery for short lived nodes.

So how about I revert the clap changes but keep the separate receive?

rklaehn: Ok, reverted the changes completely for now.

You think it would be good to add this discovery detail to the quickstart? It would make it much longer, so maybe it is fine to have that in the other examples.

matheus23: Oh yeah don't do this in other examples! :D The transfer example is a special thing, as it's got a *loooot* of explanation around it in the quickstart.

> So how about I revert the clap changes but keep the separate receive?

I suggest you look at changing the website's quickstart and see how it feels to explain that version of the code there. If it feels good, then let's go for it :) (Also I'd like to change both the website's quickstart and this in tandem, if possible)

> The reason I split the send and receive path is because I dislike the (ephemeral) receive side publishing it's addr info to discovery. I think it is good to encourage people to use passive discovery for short lived nodes.

That's fair - that's something we don't explain in the quickstart. It might be nice to do that, though.

matheus23: > You think it would be good to add this discovery detail to the quickstart? It would make it much longer, so maybe it is fine to have that in the other examples.

Yeah I think that might be worth explaining!

PR #123, opened 2025-07-24
fix: Hide the xxx_sink variants from the public API

rklaehn:
## Description

These functions are internal and have no business being in the public API, I think

## Breaking Changes

Removes
- api::remote::Remote::fetch_sink
- api::remote::Remote::execute_push_sink
- api::remote::Remote::execute_get_sink

## Notes & open questions

Note: maybe we need to expose them at some point, but currently I don't see the need, and in any case the Sink trait is not exposed, so nobody can use them anyway at this time!

## Change checklist

- [ ] Self-review.
- [ ] Tests if relevant.

### DIFF
--- a/src/api/remote.rs
+++ b/src/api/remote.rs
@@ -499,7 +499,7 @@ impl Remote {
     /// is the aggregated number of downloaded payload bytes in the request.
     ///
     /// This will return the stats of the download.
-    pub async fn fetch_sink(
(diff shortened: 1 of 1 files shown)


Arqu (approved):

rklaehn:
> They use a custom `Sink` trait, so yeah I think so too :)

Yeah, I needed something like a sink, but without all the poll madness of the one that is in futures...

PR #128, opened 2025-07-28
deps: bump to irpc@main

Frando:
## Description

This bumps irpc to the (not yet released) main version, adapting for the breaking changes introduced in https://github.com/n0-computer/irpc/pull/46.

## Change checklist

- [ ] Self-review.
- [ ] Tests if relevant.

### DIFF
--- a/src/api.rs
+++ b/src/api.rs
@@ -19,7 +19,7 @@ use iroh::Endpoint;
 use irpc::rpc::{listen, Handler};
 use n0_snafu::SpanTrace;
 use nested_enum_utils::common_fields;
-use proto::{Request, ShutdownRequest, SyncDbRequest};
+use proto::{BlobsApi, ShutdownRequest, SyncDbRequest};
 use ref_cast::RefCast;
 use serde::{Deserialize, Serialize};
 use snafu::{Backtrace, IntoError, Snafu};
@@ -32,7 +32,7 @@ pub mod remote;
 pub mod tags;
 pub use crate::{store::util::Tag, util::temp_tag::TempTag};
 
-pub(crate) type ApiClient = irpc::Client<proto::Command, proto::Request, proto::StoreService>;
+pub(crate) type ApiClient = irpc::Client<proto::BlobsApi>;
 
 #[common_fields({
     backtrace: Option<Backtrace>,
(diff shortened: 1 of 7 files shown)


src/api.rs
```diff
@@ -281,43 +281,43 @@ impl Store {
...
-                    Request::ExportBao(msg) => local.send((msg, tx)),
-                    Request::ExportRanges(msg) => local.send((msg, tx)),
-                    Request::ExportPath(msg) => local.send((msg, tx)),
-
-                    Request::Observe(msg) => local.send((msg, tx)),
-
-                    Request::ClearProtected(msg) => local.send((msg, tx)),
-                    Request::SyncDb(msg) => local.send((msg, tx)),
-                    Request::Shutdown(msg) => local.send((msg, tx)),
+                    BlobsApi::SetTag(msg) => local.send((msg, tx)),
```
rklaehn: Was this forwarder boilerplate one of the things we can get rid of with the new features of the macro?

Just curious, perfectly reasonable to do this in a subsequent PR.

src/api/proto.rs
```diff
@@ -86,13 +86,9 @@ impl HashSpecific for CreateTagMsg {
...
 
-#[derive(Debug, Clone)]
-pub struct StoreService;
-impl irpc::Service for StoreService {}
-
-#[rpc_requests(StoreService, message = Command, alias = "Msg")]
+#[rpc_requests(message = Command, alias = "Msg")]
 #[derive(Debug, Serialize, Deserialize)]
-pub enum Request {
+pub enum BlobsApi {
```
rklaehn: Not quite sure about this rename. Request as the enum of all possible serializable requests seems fine. And calling it BlobsApi creates a bit of confusion with the api module that has the actual friendly api.

src/api.rs
Frando: Yup! Removed in the latest commit.

src/api/proto.rs
Frando: Don't mind much. Renamed back to `Request`. As most things are not public that use this it's fine. My initial rename was from seeing `irpc::Client<Request>` and thinking "maybe better to rename to something that contains `Blobs`". But as this is all private it really is fine as it is.

PR #129, opened 2025-07-28
feat: Add back builder for blobs protocol

rklaehn:
## Description

We used to have a builder in iroh-blobs. This brings it back. It does not have as many options as the previous one, but there will be more, probably.

I left the BlobsProtocol::new public for now. We can make it private if the options become too many.

## Breaking Changes

None

## Notes & open questions

Make new private right now?

## Change checklist

- [ ] Self-review.
- [ ] Tests if relevant.


rklaehn:
I was able to remove the endpoint from BlobsProtocol, so I think having just a constructor is actually fine...

PR #137, opened 2025-08-20
feat: Use reflink_or_copy for export as well

rklaehn:
## Description

Use reflink_or_copy when exporing, not just when importing.

## Breaking Changes

None

## Notes & open questions

Note: we lose progress, but for modern file systems it will be instant, so 🤷 . If you are a poor person having to live with Fat32, you can always use TryReference.

## Change checklist

- [ ] Self-review.
- [ ] Tests if relevant.

### DIFF
--- a/src/store/fs.rs
+++ b/src/store/fs.rs
@@ -1263,9 +1263,18 @@ async fn export_path_impl(
         }
         MemOrFile::File((source_path, size)) => match mode {
             ExportMode::Copy => {
-                let source = fs::File::open(&source_path)?;
-                let mut target = fs::File::create(&target)?;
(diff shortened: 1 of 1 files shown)


rklaehn:
> Looking at the implementation of `reflink_or_copy` it should be pretty easy to have our cake and eat it too: We can try to reflink, and then do the fallback ourselves, and if we fall back, we keep using `copy_with_progress`, so we have progress.

Good point. I had briefly looked at https://docs.rs/reflink-copy/latest/reflink_copy/fn.check_reflink_support.html, but it says that it just works on windows. But just trying it is better.

PR #138, opened 2025-08-21
refactor: Replace connection pool

rklaehn:
## Description

Replaces the somewhat hackish connection pool with the one from https://github.com/n0-computer/iroh-experiments/pull/36 that was battle tested more.

## Breaking Changes

None

## Notes & open questions

Q: Expose the conn pool here?

Note: There is a nice list of possible extensions, but I think it is probably best to first get the basic version in.

Extensions would be `async fn ban(node_id: NodeId, duration: Option<Duration>)`, bans the node for a time, or for as long as the conn pool lives if duration is set to None.

A way to observe pool stats so users know when to schedule new downloads without having to try.

## Change checklist

- [ ] Self-review.
- [ ] Tests if relevant.

### DIFF
--- a/src/api/downloader.rs
+++ b/src/api/downloader.rs
@@ -442,7 +442,7 @@ async fn execute_get(
                 request: request.clone(),
             })
             .await?;
-        let conn = pool.connect(provider);
+        let conn = pool.get_or_connect(provider);
         let local = remote.local_for_request(request.clone()).await?;
         if local.is_complete() {
             return Ok(());
--- a/src/api/downloader.rs
+++ b/src/api/downloader.rs
@@ -23,7 +23,10 @@ use tracing::{info, instrument::Instrument, warn};
 use super::{remote::GetConnection, Store};
 use crate::{
     protocol::{GetManyRequest, GetRequest},
-    util::sink::{Drain, IrpcSenderRefSink, Sink, TokioMpscSenderSink},
+    util::{
+        connection_pool::ConnectionPool,
+        sink::{Drain, IrpcSenderRefSink, Sink, TokioMpscSenderSink},
+    },
     BlobFormat, Hash, HashAndFormat,
 };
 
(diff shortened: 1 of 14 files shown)


src/api/downloader.rs
```diff
@@ -526,13 +442,13 @@ async fn execute_get(
                 request: request.clone(),
             })
             .await?;
-        let mut conn = pool.dial(provider);
+        let conn = pool.connect(provider);
```
matheus23: Minor nit: How about `ConnectionPool::get_or_connect`?

src/api/remote.rs
```diff
@@ -656,13 +656,15 @@ impl Remote {
...
+        conn: &Connection,
         request: GetRequest,
         mut progress: impl Sink<u64, Error = io::Error>,
     ) -> GetResult<Stats> {
         let store = self.store();
         let root = request.hash;
-        let start = crate::get::fsm::start(conn, request, Default::default());
+        // I am cloning the connection, but it's fine because the original connection or ConnectionRef stays alive
+        // for the duration of the operation.
+        let start = crate::get::fsm::start(conn.clone(), request, Default::default());
```
matheus23, line 667: Is the alternative to thread through a `ConnectionRef`? I.e. can we make `crate::get::fsm::start` take a `ConnectionRef` instead of a `Connection`?

I don't know if `get::fsm::start` is also called with normal `Connection`s. In that case, is it perhaps possible to create a `ConnectionRef` from a `Connection` without a pool?

rklaehn: I don't want fsm::start to be bound to ConnectionRef. One alternative would be to make the entire FSM generic and take a `impl Deref<Target = Connection>`. But since the FSM is a lot of types I am a bit hesitant to do this. If only rust had a way to parametrize a group of types at once.

src/api/downloader.rs
rklaehn: Called it get_or_connect after all. I was initially thinking that connect just looks nicer, but get_or_connect is more accurate.

src/api/remote.rs
rklaehn: Reason I don't want ConnectionRef in the fsm is that I might want to move the conn pool somewhere else at some point. If something like this is useful enough we could add it to iroh itself, then I guess you could do it!

src/protocol.rs
```diff
@@ -714,6 +718,73 @@ impl TryFrom<VarInt> for Closed {
     }
 }
 
+pub trait ChunkRangesExt {
```
rklaehn, line 721: I guess this thing should move to bao-tree at some point!

PR #140, opened 2025-08-26
feat: let connection pool watch for connection close

rklaehn:
## Description

Let the per-connection actor watch for connection close. When the connection is closed the conn actor will tell the main actor to remove the per-conn actor, and the next connection attempt will get a new per-connection actor.

## Breaking Changes

None

## Notes & open questions

Note: there is a very low probability that a request to get a ConnectionRef will return the closed connection. But that is fine. If the connection is closed, all currently live ConnectionRefs will turn unusable anyway, what's one more? In connected state the actor will be very fast in handing out ConnectionRef, so the queue should never be full. The queue is mostly for waiting during connect.

## Change checklist

- [ ] Self-review.
- [ ] Tests if relevant.

### DIFF
--- a/src/util/connection_pool.rs
+++ b/src/util/connection_pool.rs
@@ -4,10 +4,10 @@
 //! ALPN and [`Options`]. Then the pool will manage connections for you.
 //!
 //! Access to connections is via the [`ConnectionPool::connect`] method, which
-//! gives you access to a connection if possible.
+//! gives you access to a connection via a [`ConnectionRef`] if possible.
 //!
-//! It is important that you use the connection only in the future passed to
-//! connect, and don't clone it out of the future.
+//! It is important that you keep the [`ConnectionRef`] alive while you are using
+//! the connection.
 use std::{
     collections::{HashMap, VecDeque},
     ops::Deref,
@@ -18,10 +18,7 @@ use std::{
     time::Duration,
 };
 
-use iroh::{
-    endpoint::ConnectError,
-    Endpoint, NodeId,
-};
+use iroh::{endpoint::ConnectError, Endpoint, NodeId};
 use n0_future::{
     future::{self},
     FuturesUnordered, MaybeFuture, Stream, StreamExt,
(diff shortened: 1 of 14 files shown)


src/util/connection_pool.rs
```diff
@@ -134,25 +160,47 @@ impl Context {
...
+                .connect(node_id, &context2.alpn)
+                .await
+                .map_err(PoolConnectError::from)?;
+            if let Some(on_connect) = &context2.options.on_connected {
+                on_connect(&context2.endpoint, &conn)
+                    .await
+                    .map_err(PoolConnectError::from)?;
+            }
+            Result::<Connection, PoolConnectError>::Ok(conn)
+        };
```
matheus23: Nit: I'm still personally a fan of the pattern:
```suggestion
        let conn_fut = {
            let context = context.clone();
            async move {
                let conn = context
                    .endpoint
                    .connect(node_id, &context.alpn)
                    .await
                    .map_err(PoolConnectError::from)?;
                if let Some(on_connect) = &context.options.on_connected {
                    on_connect(&context.endpoint, &conn)
                        .await
                        .map_err(PoolConnectError::from)?;
                }
                Result::<Connection, PoolConnectError>::Ok(conn)
            }
        };
```

src/util/connection_pool.rs
```diff
@@ -134,25 +160,47 @@ impl Context {
...
+            .and_then(|r| r);
+        let conn_close = match &state {
+            Ok(conn) => {
+                let conn = conn.clone();
+                MaybeFuture::Some(async move { conn.closed().await })
             }
-        }
+            Err(e) => {
+                debug!(%node_id, "Failed to connect {e:?}, requesting shutdown");
+                tokio::time::sleep(Duration::from_secs(1)).await;
```
matheus23: Why do we sleep here?

src/util/connection_pool.rs
flub: I concur with @matheus23 !

src/util/connection_pool.rs
rklaehn: I haven't got the slightest idea. Probably wanted to see if some test fails, and forgot to remove. Thanks for noticing.

PR #142, opened 2025-08-28
feat: Provider events refactor

rklaehn:
## Description

Refactor provider events into a proper irpc protocol.

Also allow configuring for each event type if the event is sent as a notification, a proper request with answer, or not at all.

## Breaking Changes

Provider events completely changed. Other than that the changes should be minimal. You can still create a BlobsProtocol by passing None.

## Notes & open questions

Note: to review, best to start with looking at the limit example, then look at the docs.

## Change checklist

- [ ] Self-review.
- [ ] Tests if relevant.

### DIFF
--- a/examples/limit.rs
+++ b/examples/limit.rs
@@ -81,7 +81,7 @@ fn limit_by_node_id(allowed_nodes: HashSet<NodeId>) -> EventSender {
     let mask = EventMask {
         // We want a request for each incoming connection so we can accept
         // or reject them. We don't need any other events.
-        connected: ConnectMode::Request,
+        connected: ConnectMode::Intercept,
         ..EventMask::DEFAULT
     };
     let (tx, mut rx) = EventSender::channel(32, mask);
@@ -108,7 +108,7 @@ fn limit_by_hash(allowed_hashes: HashSet<Hash>) -> EventSender {
         // We want to get a request for each get request that we can answer
         // with OK or not OK depending on the hash. We do not want detailed
         // events once it has been decided to handle a request.
-        get: RequestMode::Request,
+        get: RequestMode::Intercept,
         ..EventMask::DEFAULT
     };
     let (tx, mut rx) = EventSender::channel(32, mask);
@@ -136,7 +136,7 @@ fn throttle(delay_ms: u64) -> EventSender {
     let mask = EventMask {
         // We want to get requests for each sent user data blob, so we can add a delay.
         // Other than that, we don't need any events.
-        throttle: ThrottleMode::Throttle,
+        throttle: ThrottleMode::Intercept,
         ..EventMask::DEFAULT
     };
     let (tx, mut rx) = EventSender::channel(32, mask);
@@ -190,7 +190,7 @@ fn limit_max_connections(max_connections: usize) -> EventSender {
         // based on the current connection count if we want to accept or reject.
         // We also want detailed logging of events for the get request, so we can
         // detect when the request is finished one way or another.
-        connected: ConnectMode::Request,
+        connected: ConnectMode::Intercept,
         ..EventMask::DEFAULT
     };
     let (tx, mut rx) = EventSender::channel(32, mask);
--- a/src/provider/events.rs
+++ b/src/provider/events.rs
@@ -16,6 +16,7 @@ use crate::{
     Hash,
 };
 
+/// Mode for connect events.
 #[derive(Debug, Clone, Copy, PartialEq, Eq, Default)]
 #[repr(u8)]
 pub enum ConnectMode {
@@ -25,9 +26,10 @@ pub enum ConnectMode {
     /// We get a notification for connect events.
     Notify,
     /// We get a request for connect events and can reject incoming connections.
-    Request,
+    Intercept,
 }
 
+/// Request mode for observe requests.
 #[derive(Debug, Clone, Copy, PartialEq, Eq, Default)]
 #[repr(u8)]
 pub enum ObserveMode {
@@ -37,9 +39,10 @@ pub enum ObserveMode {
     /// We get a notification for connect events.
     Notify,
     /// We get a request for connect events and can reject incoming connections.
-    Request,
+    Intercept,
 }
 
+/// Request mode for all data related requests.
 #[derive(Debug, Clone, Copy, PartialEq, Eq, Default)]
 #[repr(u8)]
 pub enum RequestMode {
@@ -49,16 +52,20 @@ pub enum RequestMode {
     /// We get a notification for each request, but no transfer events.
     Notify,
     /// We get a request for each request, and can reject incoming requests, but no transfer events.
-    Request,
+    Intercept,
     /// We get a notification for each request as well as detailed transfer events.
     NotifyLog,
     /// We get a request for each request, and can reject incoming requests.
     /// We also get detailed transfer events.
-    RequestLog,
+    InterceptLog,
     /// This request type is completely disabled. All requests will be rejected.
+    ///
+    /// This means that requests of this kind will always be rejected, whereas
+    /// None means that we don't get any events, but requests will be processed normally.
     Disabled,
 }
 
+/// Throttling mode for requests that support throttling.
 #[derive(Debug, Clone, Copy, PartialEq, Eq, Default)]
 #[repr(u8)]
 pub enum ThrottleMode {
@@ -66,15 +73,18 @@ pub enum ThrottleMode {
     #[default]
     None,
     /// We call throttle to give the event handler a way to throttle requests
-    Throttle,
+    Intercept,
 }
 
 #[derive(Debug, Clone, Copy, PartialEq, Eq, Serialize, Deserialize)]
 pub enum AbortReason {
+    /// The request was aborted because a limit was exceeded. It is OK to try again later.
     RateLimited,
+    /// The request was aborted because the client does not have permission to perform the operation.
     Permission,
 }
 
+/// Errors that can occur when sending progress updates.
 #[derive(Debug, Snafu)]
 pub enum ProgressError {
     Limit,
@@ -141,6 +151,10 @@ impl From<irpc::channel::SendError> for ProgressError {
 pub type EventResult = Result<(), AbortReason>;
 pub type ClientResult = Result<(), ProgressError>;
 
+/// Event mask to configure which events are sent to the event handler.
+///
+/// This can also be used to completely disable certain request types. E.g.
+/// push requests are disabled by default, as they can write to the local store.
 #[derive(Debug, Clone, Copy, PartialEq, Eq)]
 pub struct EventMask {
     /// Connection event mask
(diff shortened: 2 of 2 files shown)


Frando:
My branch had some mistakes, sorry for that - fixups are here: https://github.com/n0-computer/iroh-blobs/pull/144/commits/65e5e401efbf90482752da58a62fc6133ff611d0

examples/limit.rs
```diff
@@ -0,0 +1,371 @@
...
+            c.fetch_sub(1, Ordering::SeqCst);
+        }
+    }
+
+    let mask = EventMask {
+        // For each get request, we want to get a request so we can decide
+        // based on the current connection count if we want to accept or reject.
+        // We also want detailed logging of events for the get request, so we can
+        // detect when the request is finished one way or another.
+        get: RequestMode::RequestLog,
```
dignifiedquire: I am confused, I thought this limits connections, but you are checking on each get request?

examples/limit.rs
```diff
@@ -0,0 +1,371 @@
+/// Example how to limit blob requests by hash and node id, and to add
```
dignifiedquire, line 1: this is  very cool overall!

src/provider/events.rs
```diff
@@ -0,0 +1,684 @@
...
+        GetManyRequest, GetRequest, ObserveRequest, PushRequest, ERR_INTERNAL, ERR_LIMIT,
+        ERR_PERMISSION,
+    },
+    provider::{events::irpc_ext::IrpcClientExt, TransferStats},
+    Hash,
+};
+
+#[derive(Debug, Clone, Copy, PartialEq, Eq, Default)]
+#[repr(u8)]
+pub enum ConnectMode {
```
dignifiedquire, line 22: maybe add top level doc comments to these  enums?

src/provider/events.rs
```diff
@@ -0,0 +1,684 @@
...
+    /// We don't get notification of connect events at all.
+    #[default]
+    None,
+    /// We get a notification for connect events.
+    Notify,
+    /// We get a request for connect events and can reject incoming connections.
+    Request,
```
dignifiedquire: I wonder if something like `Interrupt` would make more sense, to signal how this works (just an idea)

src/provider/events.rs
```diff
@@ -0,0 +1,684 @@
...
+    Notify,
+    /// We get a request for each request, and can reject incoming requests, but no transfer events.
+    Request,
+    /// We get a notification for each request as well as detailed transfer events.
+    NotifyLog,
+    /// We get a request for each request, and can reject incoming requests.
+    /// We also get detailed transfer events.
+    RequestLog,
+    /// This request type is completely disabled. All requests will be rejected.
+    Disabled,
```
dignifiedquire, line 65: It's a bit hard to understand the difference between `None` and `Disabled` I think

examples/limit.rs
rklaehn: You're right of course. The counter limits concurrent get requests.

src/provider/events.rs
```diff
@@ -0,0 +1,684 @@
...
+}
+
+#[derive(Debug, Clone, Copy, PartialEq, Eq, Default)]
+#[repr(u8)]
+pub enum ThrottleMode {
+    /// We don't get these kinds of events at all
+    #[default]
+    None,
+    /// We call throttle to give the event handler a way to throttle requests
+    Throttle,
```
dignifiedquire: shouldn't this be called `Request` to match the structure in the others?

src/provider/events.rs
```diff
@@ -0,0 +1,684 @@
...
+}
+
+#[derive(Debug, Clone, Copy, PartialEq, Eq, Serialize, Deserialize)]
+pub enum AbortReason {
+    RateLimited,
```
dignifiedquire, line 82: needs docs

src/provider/events.rs
```diff
@@ -0,0 +1,684 @@
...
+    }
+}
+
+pub type EventResult = Result<(), AbortReason>;
+pub type ClientResult = Result<(), ProgressError>;
+
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub struct EventMask {
+    /// Connection event mask
+    pub connected: ConnectMode,
```
dignifiedquire, line 161: you could implement `From<ConnectMode> for EventMask` and then I could write sth like `Channel::new(32, ConnectMode::Request)` if you make the function take `impl Into<EventMask>`

src/provider/events.rs
```diff
@@ -0,0 +1,684 @@
...
+        Self { updates, throttle }
+    }
+
+    /// A request tracker that doesn't track anything.
+    pub const NONE: Self = Self {
+        updates: RequestUpdates::None,
+        throttle: None,
+    };
+
+    /// Transfer for index `index` started, size `size`
```
dignifiedquire: size in bytes?

src/provider/events.rs
```diff
@@ -0,0 +1,684 @@
...
+                }
+                .into(),
+            )
+            .await?;
+        }
+        Ok(())
+    }
+
+    /// Transfer progress for the previously reported blob, end_offset is the new end offset in bytes.
+    pub async fn transfer_progress(&mut self, len: u64, end_offset: u64) -> ClientResult {
```
dignifiedquire, line 273: maybe normalize to `size` or `len` overall?

src/provider/events.rs
```diff
@@ -0,0 +1,684 @@
...
+    ) -> (Self, tokio::sync::mpsc::Receiver<ProviderMessage>) {
+        let (tx, rx) = tokio::sync::mpsc::channel(capacity);
+        (Self::new(tx, mask), rx)
+    }
+
+    /// Log request events at trace level.
+    pub fn tracing(&self, mask: EventMask) -> Self {
+        use tracing::trace;
+        let (tx, mut rx) = tokio::sync::mpsc::channel(32);
+        n0_future::task::spawn(async move {
```
dignifiedquire, line 336: no task tracking?

src/provider/events.rs
```diff
@@ -0,0 +1,684 @@
...
+            ) {
+                n0_future::task::spawn(async move {
+                    while let Ok(Some(update)) = rx.recv().await {
+                        trace!(%connection_id, %request_id, "{update:?}");
+                    }
+                });
+            }
+            while let Some(msg) = rx.recv().await {
+                match msg {
+                    ProviderMessage::ClientConnected(_) => todo!(),
```
dignifiedquire: what about this?

src/provider/events.rs
```diff
@@ -0,0 +1,684 @@
...
+            match self.mask.connected {
+                ConnectMode::None => {}
+                ConnectMode::Notify => client.notify(Notify(f())).await?,
+                ConnectMode::Request => client.rpc(f()).await??,
+            }
+        };
+        Ok(())
+    }
+
+    /// A new client has been connected.
```
dignifiedquire: wrong docs

src/provider/events.rs
```diff
@@ -0,0 +1,684 @@
...
+        ProviderMessage: From<WithChannels<RequestReceived<Req>, ProviderProto>>,
+        RequestReceived<Req>: Channels<
+            ProviderProto,
+            Tx = oneshot::Sender<EventResult>,
+            Rx = mpsc::Receiver<RequestUpdate>,
+        >,
+        ProviderProto: From<Notify<RequestReceived<Req>>>,
+        ProviderMessage: From<WithChannels<Notify<RequestReceived<Req>>, ProviderProto>>,
+        Notify<RequestReceived<Req>>:
+            Channels<ProviderProto, Tx = NoSender, Rx = mpsc::Receiver<RequestUpdate>>,
```
dignifiedquire, line 449: 😭

src/provider/events.rs
```diff
@@ -0,0 +1,684 @@
...
+    /// A new client connected to the provider. Notify variant.
+    #[rpc(tx = NoSender)]
+    ClientConnectedNotify(Notify<ClientConnected>),
+
+    /// A client disconnected from the provider.
+    #[rpc(tx = NoSender)]
+    ConnectionClosed(ConnectionClosed),
+
+    #[rpc(rx = mpsc::Receiver<RequestUpdate>, tx = oneshot::Sender<EventResult>)]
+    /// A new get request was received from the provider.
```
dignifiedquire, line 535: doc comments below are all copy & paste

src/provider/events.rs
rklaehn: Interrupt would imply that it always interrupts. Intercept?

src/provider/events.rs
```diff
@@ -0,0 +1,684 @@
...
+        pub connection_id: u64,
+        /// The request id. There is a new id for each request.
+        pub request_id: u64,
+        /// The request
+        pub request: R,
+    }
+
+    /// Request to throttle sending for a specific request.
+    #[derive(Debug, Serialize, Deserialize)]
+    pub struct Throttle {
```
dignifiedquire, line 602: so this is request throtteling? should there be connection throtteling as well?

src/provider/events.rs
rklaehn: Yeah, I had this called Throttle since at some point I had Throttle return a (), so you could not actually stop a request during throttling. But now I made it possible, so it should be named consistently.

src/provider/events.rs
rklaehn: Sigh, fixed it already, but not on the right branch...

src/provider/events.rs
rklaehn: Yeah, DRYing stuff with lots of constraints is so much fun that it is often just not worth it...

examples/limit.rs
rklaehn: I modified the example to actually limit connections - makes it much simpler.

src/provider/events.rs
dignifiedquire: I like `Intercept`, it signals more strongly that it does potentially stop the ongoing flow

src/provider/events.rs
rklaehn: What would that be? You can limit the number of connections already, but throttling here is in relation to the amount of data being sent. It would be up to the throttler (sp?) to do some bookkeeping to make sure there is a certain amount of bandwith per connection. The throttle event contains the required information. `connection_id` is in here, and even if there are multiple connections to the same remote, you can map it with the info you get in ClientConnected.

Would be a more complex piece of code though, so I thought I leave it as an exercise for the reader...

dignifiedquire: makes sense, I was just thinking out loudly for the most part

src/provider/events.rs
rklaehn: Not sure how useful this would be. EventMask is a **mask**, so what if you want multiple things to be enabled? Implement | (or)? Then what is Notify | Request? Request? I think it would be confusing.

dignifiedquire: implement `or` is what I would do, yeah

src/provider/events.rs
rklaehn: Added a comment. None is no events but proceed, Disabled is no events but fail.

src/provider/events.rs
rklaehn: Yeah, I was too lazy to make this a full fledged actor. Not sure if it is worth it here.

PR #143, opened 2025-09-03
chore: Feature gate redb based store.

rklaehn:
## Description

Feature gate redb based store. It is still enabled by default. Many tests and examples require this to be enabled, but I think that is fine, since we run both with all and with no default features.

## Breaking Changes

None

## Notes & open questions

None

## Change checklist

- [ ] Self-review.
- [ ] Tests if relevant.

### DIFF
--- a/src/api/blobs.rs
+++ b/src/api/blobs.rs
@@ -144,7 +144,7 @@ impl Blobs {
     /// clears the protections before.
     ///
     /// Users should rely only on garbage collection for blob deletion.
-    #[allow(dead_code)]
+    #[cfg(feature = "fs-store")]
     pub(crate) async fn delete_with_opts(&self, options: DeleteOptions) -> RequestResult<()> {
         trace!("{options:?}");
         self.client.rpc(options).await??;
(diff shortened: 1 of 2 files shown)


Cargo.toml
```diff
@@ -64,7 +64,8 @@ iroh = { version = "0.91.1", features = ["discovery-local-network"]}
 [features]
 hide-proto-docs = []
 metrics = []
-default = ["hide-proto-docs"]
+default = ["hide-proto-docs", "fs-store"]
+fs-store = ["dep:redb", "dep:reflink-copy"]
```
dignifiedquire, line 68: maybe add some docs either here or to the readme/doc comments in lib.rs?

src/api/blobs.rs
```diff
@@ -144,13 +144,15 @@ impl Blobs {
     /// clears the protections before.
     ///
     /// Users should rely only on garbage collection for blob deletion.
+    #[allow(dead_code)]
```
dignifiedquire: why allow dead_code, shouldn't this be dependent on the feature instead?

src/api/blobs.rs
```diff
@@ -510,6 +512,7 @@ impl Blobs {
         }
     }
 
+    #[allow(dead_code)]
```
dignifiedquire, line 515: what's with this one?

PR #147, opened 2025-09-11
feat: abstract over stream types on provide and get side

rklaehn:
## Description

This PR introduces abstract versions of iroh::endpoint::{SendStream, RecvStream} and modifies the provide side and get side implementation to use these abstract streams instead of directly using iroh::endpoint streams.

This is necessary for wrapping the streams into a transformation such as compression, see the discussion in https://github.com/n0-computer/sendme/pull/93 .

The compression example shows how streams can be wrapped into compression/decompression to create a derived protocol with a different ALPN that is identical to the blobs protocol except for compression.

## Breaking Changes

iroh::endpoint::SendStream and iroh::endpoint::RecvStream are replaced with the traits iroh_blobs::util::SendStream and iroh_blobs::util::RecvStream in the get FSM and in the provider side API.

## Change checklist

- [ ] Self-review.
- [ ] Tests if relevant.

### DIFF
--- a/src/util/stream.rs
+++ b/src/util/stream.rs
@@ -17,10 +17,7 @@ pub trait SendStream: Send {
     /// This method is not cancellation safe. Even if this does not resolve, some bytes may have been written when previously polled.
     fn send_bytes(&mut self, bytes: Bytes) -> impl Future<Output = io::Result<()>> + Send;
     /// Send that sends a fixed sized buffer.
-    fn send<const L: usize>(
-        &mut self,
-        buf: &[u8; L],
-    ) -> impl Future<Output = io::Result<()>> + Send;
+    fn send(&mut self, buf: &[u8]) -> impl Future<Output = io::Result<()>> + Send;
     /// Sync the stream. Not needed for iroh, but needed for intermediate buffered streams such as compression.
     fn sync(&mut self) -> impl Future<Output = io::Result<()>> + Send;
     /// Reset the stream with the given error code.
@@ -41,8 +38,8 @@ pub trait RecvStream: Send {
     ///
     /// Note that this is different from `recv_bytes`, which will return fewer bytes if the stream ends.
     fn recv_bytes_exact(&mut self, len: usize) -> impl Future<Output = io::Result<Bytes>> + Send;
-    /// Receive exactly `L` bytes from the stream, directly into a `[u8; L]`.
-    fn recv<const L: usize>(&mut self) -> impl Future<Output = io::Result<[u8; L]>> + Send;
+    /// Receive exactly `target.len()` bytes from the stream.
+    fn recv_exact(&mut self, target: &mut [u8]) -> impl Future<Output = io::Result<()>> + Send;
     /// Stop the stream with the given error code.
     fn stop(&mut self, code: VarInt) -> io::Result<()>;
     /// Get the stream id.
@@ -54,7 +51,7 @@ impl SendStream for iroh::endpoint::SendStream {
         Ok(self.write_chunk(bytes).await?)
     }
 
-    async fn send<const L: usize>(&mut self, buf: &[u8; L]) -> io::Result<()> {
+    async fn send(&mut self, buf: &[u8]) -> io::Result<()> {
         Ok(self.write_all(buf).await?)
     }
 
@@ -100,14 +97,12 @@ impl RecvStream for iroh::endpoint::RecvStream {
         Ok(buf.into())
     }
 
-    async fn recv<const L: usize>(&mut self) -> io::Result<[u8; L]> {
-        let mut buf = [0; L];
-        self.read_exact(&mut buf).await.map_err(|e| match e {
+    async fn recv_exact(&mut self, buf: &mut [u8]) -> io::Result<()> {
+        self.read_exact(buf).await.map_err(|e| match e {
             ReadExactError::FinishedEarly(0) => io::Error::new(io::ErrorKind::UnexpectedEof, ""),
             ReadExactError::FinishedEarly(_) => io::Error::new(io::ErrorKind::InvalidData, ""),
             ReadExactError::ReadError(e) => e.into(),
-        })?;
-        Ok(buf)
+        })
     }
 
     fn stop(&mut self, code: VarInt) -> io::Result<()> {
@@ -128,8 +123,8 @@ impl<R: RecvStream> RecvStream for &mut R {
         self.deref_mut().recv_bytes_exact(len).await
     }
 
-    async fn recv<const L: usize>(&mut self) -> io::Result<[u8; L]> {
-        self.deref_mut().recv::<L>().await
+    async fn recv_exact(&mut self, buf: &mut [u8]) -> io::Result<()> {
+        self.deref_mut().recv_exact(buf).await
     }
 
     fn stop(&mut self, code: VarInt) -> io::Result<()> {
@@ -146,7 +141,7 @@ impl<W: SendStream> SendStream for &mut W {
         self.deref_mut().send_bytes(bytes).await
     }
 
-    async fn send<const L: usize>(&mut self, buf: &[u8; L]) -> io::Result<()> {
+    async fn send(&mut self, buf: &[u8]) -> io::Result<()> {
         self.deref_mut().send(buf).await
     }
 
@@ -174,8 +169,15 @@ pub struct AsyncReadRecvStream<R>(R);
 /// `AsyncRead + Unpin + Send`, you can implement these additional methods and wrap the result
 /// in an `AsyncReadRecvStream` to get a `RecvStream` that reads from the underlying `AsyncRead`.
 pub trait AsyncReadRecvStreamExtra: Send {
+    /// Get a mutable reference to the inner `AsyncRead`.
+    ///
+    /// Getting a reference is easier than implementing all methods on `AsyncWrite` with forwarders to the inner instance.
     fn inner(&mut self) -> &mut (impl AsyncRead + Unpin + Send);
+    /// Stop the stream with the given error code.
     fn stop(&mut self, code: VarInt) -> io::Result<()>;
+    /// A local unique identifier for the stream.
+    ///
+    /// This allows distinguishing between streams, but once the stream is closed, the id may be reused.
     fn id(&self) -> u64;
 }
 
@@ -209,10 +211,9 @@ impl<R: AsyncReadRecvStreamExtra> RecvStream for AsyncReadRecvStream<R> {
         Ok(res.into())
     }
 
-    async fn recv<const L: usize>(&mut self) -> io::Result<[u8; L]> {
-        let mut res = [0; L];
-        self.0.inner().read_exact(&mut res).await?;
-        Ok(res)
+    async fn recv_exact(&mut self, buf: &mut [u8]) -> io::Result<()> {
+        self.0.inner().read_exact(buf).await?;
+        Ok(())
     }
 
     fn stop(&mut self, code: VarInt) -> io::Result<()> {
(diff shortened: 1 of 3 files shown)


src/util/stream.rs
```diff
@@ -0,0 +1,433 @@
...
+use iroh_io::AsyncStreamReader;
+use serde::{de::DeserializeOwned, Serialize};
+use tokio::io::{AsyncRead, AsyncReadExt, AsyncWrite, AsyncWriteExt};
+
+/// An abstract `iroh::endpoint::SendStream`.
+pub trait SendStream: Send {
+    /// Send bytes to the stream. This takes a `Bytes` because iroh can directly use them.
+    fn send_bytes(&mut self, bytes: Bytes) -> impl Future<Output = io::Result<()>> + Send;
+    /// Send that sends a fixed sized buffer.
+    fn send<const L: usize>(
```
Frando: Why the `const L` instead of taking `&[u8]`?

src/util/stream.rs
```diff
@@ -0,0 +1,433 @@
...
+        &mut self,
+        buf: &[u8; L],
+    ) -> impl Future<Output = io::Result<()>> + Send;
+    /// Sync the stream. Not needed for iroh, but needed for intermediate buffered streams such as compression.
+    fn sync(&mut self) -> impl Future<Output = io::Result<()>> + Send;
```
Frando, line 22: I would have expected this to be called `flush`, or is it different from `flush` in `AsyncWrite`? If so, can you add some more docs what *sync* means here?

examples/compression.rs
```diff
@@ -0,0 +1,224 @@
...
+
+    struct SendStream(Lz4Encoder<iroh::endpoint::SendStream>);
+
+    impl SendStream {
+        pub fn new(inner: iroh::endpoint::SendStream) -> AsyncWriteSendStream<Self> {
+            AsyncWriteSendStream::new(Self(Lz4Encoder::new(inner)))
+        }
+    }
+
+    impl SendStreamSpecific for SendStream {
```
dignifiedquire: `SendStreamSpecific` is a bit of an unfortunate name

src/util/stream.rs
```diff
@@ -0,0 +1,433 @@
...
+    /// Reset the stream with the given error code.
+    fn reset(&mut self, code: VarInt) -> io::Result<()>;
+    /// Wait for the stream to be stopped, returning the error code if it was.
+    fn stopped(&mut self) -> impl Future<Output = io::Result<Option<VarInt>>> + Send;
```
Frando, line 26: `quinn::SendStream::stopped` returns a future that is `'static` in addtion to `Send`, which makes it easier to e.g. put the future into a combinator. So maybe add that bound in this trait too? Not fully sure though if that is too much of a restriction though for implementors like compression.

src/util/stream.rs
```diff
@@ -0,0 +1,433 @@
...
+};
+
+use bytes::Bytes;
+use iroh::endpoint::{ReadExactError, VarInt};
+use iroh_io::AsyncStreamReader;
+use serde::{de::DeserializeOwned, Serialize};
+use tokio::io::{AsyncRead, AsyncReadExt, AsyncWrite, AsyncWriteExt};
+
+/// An abstract `iroh::endpoint::SendStream`.
+pub trait SendStream: Send {
```
Frando, line 14: Maybe add `id()` method here too?

src/util/stream.rs
dignifiedquire: why not make it implement `tokio::io::Write` instead?

src/util/stream.rs
```diff
@@ -0,0 +1,433 @@
...
+    /// Reset the stream with the given error code.
+    fn reset(&mut self, code: VarInt) -> io::Result<()>;
+    /// Wait for the stream to be stopped, returning the error code if it was.
+    fn stopped(&mut self) -> impl Future<Output = io::Result<Option<VarInt>>> + Send;
+}
+
+/// An abstract `iroh::endpoint::RecvStream`.
+pub trait RecvStream: Send {
+    /// Receive up to `len` bytes from the stream, directly into a `Bytes`.
+    fn recv_bytes(&mut self, len: usize) -> impl Future<Output = io::Result<Bytes>> + Send;
```
dignifiedquire, line 34: why not make it implement `tokio::io::Read` instead? and provide this as a helper method?

src/util/stream.rs
```diff
@@ -0,0 +1,433 @@
...
+use bytes::Bytes;
+use iroh::endpoint::{ReadExactError, VarInt};
+use iroh_io::AsyncStreamReader;
+use serde::{de::DeserializeOwned, Serialize};
+use tokio::io::{AsyncRead, AsyncReadExt, AsyncWrite, AsyncWriteExt};
+
+/// An abstract `iroh::endpoint::SendStream`.
+pub trait SendStream: Send {
+    /// Send bytes to the stream. This takes a `Bytes` because iroh can directly use them.
+    fn send_bytes(&mut self, bytes: Bytes) -> impl Future<Output = io::Result<()>> + Send;
```
Frando, line 18: Maybe add the note on cancellation safety from [`quinn::SendStream::write_chunk`](https://docs.rs/quinn/latest/quinn/struct.SendStream.html#method.write_chunk) here too:

> ### Cancel safety
>
> This method is not cancellation safe. Even if this does not resolve, some bytes may have been written when previously polled.

src/util/stream.rs
```diff
@@ -0,0 +1,433 @@
...
+    fn stop(&mut self, code: VarInt) -> io::Result<()> {
+        self.0.stop(code)
+    }
+
+    fn id(&self) -> u64 {
+        self.0.id()
+    }
+}
+
+pub trait RecvStreamSpecific: Send {
```
dignifiedquire: this naming is confusing, what exactly are you trying to say with `Specific`?

src/util/stream.rs
```diff
@@ -0,0 +1,433 @@
...
+    /// Receive up to `len` bytes from the stream, directly into a `Bytes`.
+    fn recv_bytes(&mut self, len: usize) -> impl Future<Output = io::Result<Bytes>> + Send;
+    /// Receive exactly `len` bytes from the stream, directly into a `Bytes`.
+    ///
+    /// This will return an error if the stream ends before `len` bytes are read.
+    ///
+    /// Note that this is different from `recv_bytes`, which will return fewer bytes if the stream ends.
+    fn recv_bytes_exact(&mut self, len: usize) -> impl Future<Output = io::Result<Bytes>> + Send;
+    /// Receive exactly `L` bytes from the stream, directly into a `[u8; L]`.
+    fn recv<const L: usize>(&mut self) -> impl Future<Output = io::Result<[u8; L]>> + Send;
```
Frando: Maybe call this `recv_exact`?

src/util/stream.rs
Frando: I found the naming of this trait a bit unintuitive. Ideas I came up with are`RecvStreamWithInner` or `AsyncReadRecv`, not sure though yet. Same for SendStreamSpecific then.

src/util/stream.rs
```diff
@@ -0,0 +1,433 @@
...
+        max_size: usize,
+    ) -> io::Result<(T, usize)> {
+        let data = self.recv_bytes(max_size).await?;
+        self.expect_eof().await?;
+        let value = postcard::from_bytes(&data)
+            .map_err(|e| io::Error::new(io::ErrorKind::InvalidData, e))?;
+        Ok((value, data.len()))
+    }
+
+    async fn read_length_prefixed<T: DeserializeOwned>(
```
Frando, line 381: Yeah these are great. We should move all this with some cleanup and better docs to an `iroh-util` crate I think. It will be useful for many people working with iroh send/recv streams even independently of the trait abstraction.

examples/compression.rs
rklaehn: Yeah, that one definitely needs a better name. It's a helper for when you have an AsyncWrite and want to add the remaining stuff, so name suggestions welcome!

src/util/stream.rs
rklaehn: I tried doing a

```rust
impl<T: SendStream> tokio::io::AsyncWrite for T
```

But then you get a collision when you do the reference stuff

```rust
impl<T: SendStream> SendStream for &mut T
```

So you can either make this trait convenient to use on its own, or make it convenient to adapt to the tokio world, but not both.

src/util/stream.rs
rklaehn: Same as above, but most notably I want the ability to return a Bytes without copying, and if you add a helper method you have to copy.

iroh::endpoint::RecvStream has fns that produce a Bytes, blobs needs Bytes, but if you go through tokio::io::AsyncRead you work with &mut [u8] and lose the Bytes / must copy. Unless there is a trick I am not aware of.

src/util/stream.rs
rklaehn: Same as above. It made sense at some point, but that was before a dozen things I tried that didn't work.

This is a helper trait to impl the iroh_blobs::util::RecvStream if you have an AsyncRead, so better names welcome.

But this is also not the main trait. To see how this is used, see the compression.rs example.

src/util/stream.rs
rklaehn: Since these traits are hopeless, they will never be dynable, I thought I might as well use a const generic since it is convenient. But sure, could change this.

In case of the recv version the code just reads nicer if you return a fixed size slice instead of taking a &mut [u8], and for returning a fixed size slice you don't have to allocate.

src/util/stream.rs
rklaehn: AsyncReadRecv is the best suggestion so far...

src/util/stream.rs
rklaehn: For iroh::endpoint::SendStream there is no flush. For the adapter to AsyncWrite, it *does* call flush!

```rust
impl<W: SendStreamSpecific> SendStream for AsyncWriteSendStream<W> {
    ...
    async fn sync(&mut self) -> io::Result<()> {
        self.0.inner().flush().await
    }
    ...
}
```

src/util/stream.rs
rklaehn: iroh::endpoint::SendStream::stopped is an async fn that doesn't explicitly promise to be 'static.

So if we wanted to do this we would have to modify iroh I think.

https://docs.rs/iroh/latest/iroh/endpoint/struct.SendStream.html#method.stopped

src/util/stream.rs
rklaehn: I am somewhat hesitant to have yet another crate, but I guess maybe we have to do it even before iroh 1.0?

src/util/stream.rs
Frando: Ah right, this ia changed in quinn main but not yet in the iroh-quinn version used in the latest iroh release.

src/util/stream.rs
dignifiedquire: https://docs.rs/tokio/latest/tokio/io/trait.AsyncRead.html uses `ReadBuf` which I believe can actually avoid this issue?

dignifiedquire: but it's tricky and involves `unsafe` in some cases I think

examples/compression.rs
```diff
@@ -72,7 +73,7 @@ mod lz4 {
         }
     }
 
-    impl SendStreamSpecific for SendStream {
+    impl AsyncWriteSendStreamExtra for SendStream {
```
dignifiedquire, line 76: maybe `AsyncWriteSendStreamExt` ? for `Extension`?

rklaehn: Hm, ...Ext traits for me are specifically for the "extension method" hack.

trait FooExt { ... }

impl<T: Foo> FooExt for T {}

src/util/stream.rs
rklaehn: Yeah, I would rather have Bytes -> Bytes instead of Bytes -> complex rube goldberg machine involving ReadBuf and unsafe -> Bytes...

src/util/stream.rs
```diff
@@ -0,0 +1,458 @@
...
+}
+
+#[derive(Debug)]
+pub struct AsyncReadRecvStream<R>(R);
+
+/// This is a helper trait to work with [`AsyncReadRecvStream`]. If you have an
+/// `AsyncRead + Unpin + Send`, you can implement these additional methods and wrap the result
+/// in an `AsyncReadRecvStream` to get a `RecvStream` that reads from the underlying `AsyncRead`.
+pub trait AsyncReadRecvStreamExtra: Send {
+    fn inner(&mut self) -> &mut (impl AsyncRead + Unpin + Send);
```
dignifiedquire, line 175: some docs on these methods would be nice

Issue #172, opened 2025-09-28
Only smart pointers should implement Deref


matthiasbeyer:
Hi.

I found in your API that `BlobsProtocol: Deref<Store>` and (even more strange) `BlobsProtocol: Deref<Blobs>`.

The [Rust API guidelines state](https://rust-lang.github.io/api-guidelines/predictability.html?highlight=Deref#only-smart-pointers-implement-deref-and-derefmut-c-deref) that only smart pointers should be implementing `Deref`.

I think it would be a good idea to remove these `Deref` implementations (and possibly others) for API clearity.

What do you think?

rklaehn:
I will have to think about it. I saw Deref used for convenience many, many times, so I wasn't aware that it is supposed to be only used for smart pointers.

PR #173, opened 2025-09-28
Remove impl Deref<Blobs> for Store

matthiasbeyer:
Part of https://github.com/n0-computer/iroh-blobs/issues/172

## Description

As per rust API guidelines [0], only smart pointers should implement `Deref`.
So, we remove the `impl Deref<Blobs> for Store` in this patch.

[0]: https://rust-lang.github.io/api-guidelines/predictability.html?highlight=Deref#only-smart-pointers-implement-deref-and-derefmut-c-deref

## Breaking Changes

The `impl Deref<Blobs> for Store` was removed. For all calls to `Store` functions that were referred to the underlying `Blobs` instance, use `Store::blobs()` to get a `&Blobs`.

## Change checklist

- [x] Self-review.
- [ ] Tests if relevant.

### DIFF
--- a/src/api.rs
+++ b/src/api.rs
@@ -12,7 +12,7 @@
 //!
 //! You can also [`connect`](Store::connect) to a remote store that is listening
 //! to rpc requests.
-use std::{io, net::SocketAddr, ops::Deref};
+use std::{io, net::SocketAddr};
 
 use bao_tree::io::EncodeError;
 use iroh::Endpoint;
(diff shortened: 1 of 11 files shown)


rklaehn:
Thanks for the PR. Not quite sure how to best do this, but I think you are right that we are overusing deref.

dignifiedquire:
I don't think we will do this for now

PR #174, opened 2025-09-30
feat: upgrade database to redb v3 compatible format

Frando:
## Description

The recently released redb v3 has incompatible changes to the database format. The recommended upgrade procedure is to upgrade the database with `Database::upgrade` on redb v2.6+. The upgrade has to be done on redb v2, not v3. See https://github.com/cberner/redb/blob/master/CHANGELOG.md#removes-support-for-file-format-v2.

This updates redb to 2.6 and performs the upgrade when opening the database. Calling upgrade on an already upgraded database is a no-op.

With this merged in the next release, we can then update redb to v3 in the version after.

## Change checklist

- [ ] Self-review.
- [ ] Tests if relevant.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -42,7 +42,7 @@ genawaiter = { version = "0.99.1", features = ["futures03"] }
 iroh-base = "0.92"
 irpc = { version = "0.8.0", features = ["rpc", "quinn_endpoint_setup", "spans", "stream", "derive"], default-features = false }
 iroh-metrics = { version = "0.35" }
-redb = { version = "=2.4", optional = true }
+redb = { version = "2.6.3", optional = true }
 reflink-copy = { version = "0.1.24", optional = true }
 
 [dev-dependencies]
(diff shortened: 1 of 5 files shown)


rklaehn (commented):
The code seems fine. Have you tried it out with a real db?

Cargo.toml
```diff
@@ -42,7 +42,7 @@ genawaiter = { version = "0.99.1", features = ["futures03"] }
 iroh-base = "0.92"
 irpc = { version = "0.8.0", features = ["rpc", "quinn_endpoint_setup", "spans", "stream", "derive"], default-features = false }
 iroh-metrics = { version = "0.35" }
-redb = { version = "=2.4", optional = true }
+redb = { version = "2.6.3", optional = true }
```
rklaehn, line 45: Why is this version 2.x and later we call it v3?

Frando: From redb 2.6 `Database::upgrade` exists and changes the database format into a forward-compatible format that works both with redb v2.6+ and redb v3. The upgrade code is only part of redb v2, it was removed in v3, which only works with the upgraded format. So we have to do the upgrade with redb v2.6, and can then upgrade to v3 in the next release after the next.

PR #176, opened 2025-10-02
docs: add example for creating & fetching a collection

b5:
## Change checklist

- [ ] Self-review.
- [ ] Tests if relevant.

### DIFF
--- a/examples/transfer-collection.rs
+++ b/examples/transfer-collection.rs
@@ -80,10 +80,7 @@ impl Node {
 
         let collection_items = collection_items
             .iter()
-            .map(|(name, tag)| {
-                let hash = tag.hash().clone();
-                (name.to_string(), hash)
-            })
+            .map(|(name, tag)| (name.to_string(), *tag.hash()))
             .collect::<Vec<_>>();
 
         let collection = Collection::from_iter(collection_items);
@@ -93,7 +90,7 @@ impl Node {
         Ok(*tt.hash())
     }
 
-    /// retrive an entire collection from a given hash and provider
+    /// retrieve an entire collection from a given hash and provider
     async fn get_collection(&self, hash: Hash, provider: NodeAddr) -> Result<()> {
         let req = HashAndFormat::hash_seq(hash);
         let addrs = Shuffled::new(vec![provider.node_id]);


examples/transfer-collection.rs
```diff
@@ -0,0 +1,126 @@
...
+struct Node {
+    store: Store,
+    /// Router with the blobs protocol registered, to accept blobs requests.
+    /// We can always get the endpoint with router.endpoint()
+    router: Router,
+}
+
+impl Node {
+    async fn new() -> Result<Self> {
+        let endpoint = Endpoint::builder().bind().await?;
```
b5: I'm skipping `.discovery_n0()` here and using `add_node_addr` directly on the endpoint in the `Node::get_collection` method below. Largely to take a highligher to the `add_node_addr` discussion we've been having.

@rklaehn: the store has a connection pool, but I can't seem to get at it? Seems the connection pool _relies_ on me being able to plumb addresses into the endpoint, is that right?

examples/transfer-collection.rs
```diff
@@ -0,0 +1,126 @@
...
+        let collection = collection.store(&self.store).await?;
+        let hash = collection.hash().clone();
+        self.store.tags().create(collection).await?;
+        Ok(hash)
+    }
+
+    /// retrive an entire collection from a given hash and provider
+    async fn get_collection(&self, hash: Hash, provider: NodeAddr) -> Result<()> {
+        self.router.endpoint().add_node_addr(provider.clone())?;
+        let req = HashAndFormat::hash_seq(hash);
```
b5, line 95: This magical incantation to feed to `download` as a `req` param was super tough to figure out. Maybe we just handle this with examples.

examples/transfer-collection.rs
```diff
@@ -0,0 +1,126 @@
...
+        let addrs = Shuffled::new(vec![provider.node_id]);
```
b5, line 96: I was surprised while working with store.downloader().download()` that the `ContentDiscovery` trait worked in `NodeId` and not `NodeAddr`, is there a reason why?

Second: would it be possible to set it up so I can pass a `Vec<NodeId>` / `Vec<NodeAddr>`, and call `.into()` on it to get a `Shuffled` content discovery? That might help remove the concept of Content Discovery entirely from what we'd pass to download.

examples/transfer-collection.rs
```diff
@@ -0,0 +1,126 @@
...
+        self.store
+            .downloader(self.router.endpoint())
```
b5, line 98: I'm super confused about why I have to re-pass the endpoint to the store here to create a downloader. I guess I'm supposed to make a downloader & store it on the `Node` struct I've defined here?

It generally feels like a higher-level API is missing that just falls back to a `DefaultDownloader`

examples/transfer-collection.rs
```diff
@@ -0,0 +1,146 @@
...
+    }
+
+    /// creates a collection from a given set of named blobs, adds it to the local store
+    /// and returns the hash of the collection.
+    async fn create_collection(&self, named_blobs: Vec<(&str, Vec<u8>)>) -> Result<Hash> {
+        let mut collection_items: HashMap<&str, TempTag> = HashMap::new();
+
+        let tx = self.store.batch().await?;
+        for (name, data) in named_blobs {
+            let tmp_tag = tx.add_slice(data).await?;
```
rklaehn: Not the end of the world, but since you got a Vec<u8> already you can use [add_bytes](https://docs.rs/iroh-blobs/latest/iroh_blobs/api/blobs/struct.Batch.html#method.add_bytes) and avoid a copy. Vec<u8> impls Into<Bytes>, and this is done without copying.

examples/transfer-collection.rs
```diff
@@ -0,0 +1,146 @@
...
+                let hash = tag.hash().clone();
+                (name.to_string(), hash)
+            })
+            .collect::<Vec<_>>();
+
+        let collection = Collection::from_iter(collection_items);
+
+        let collection = collection.store(&self.store).await?;
+        let hash = collection.hash().clone();
+        self.store.tags().create(collection).await?;
```
rklaehn: This is a bug. You are creating a tag for the *hash* of the collection, but the format will be Raw, so you only protect the collection itself but not the children.

You need to use the hash_and_format of the tag.

```rust
self.store.tags().create(*collection.hash_and_format()).await
```

examples/transfer-collection.rs
rklaehn: > I was surprised while working with store.downloader().download()that theContentDiscoverytrait worked inNodeIdand notNodeAddr`, is there a reason why?

I think it is nicer to let the downloader deal only with node ids and let the node discovery figure out how to dial the nodes.

I guess you could use NodeAddr in the public API, and then just call endpoint.add_node_addr before handing it over to the lower levels. But I am very much convinced that the lower levels should not do NodeAddrs.

Any large scale content discovery mechanism will only ever work with node ids, since node ids are longer time stable than addrs.

> Second: would it be possible to set it up so I can pass a Vec<NodeId> / Vec<NodeAddr>, and call .into() on it to get a Shuffled content discovery? That might help remove the concept of Content Discovery entirely from what we'd pass to download.

I don't quite get this. You don't always want shuffled. And since ContentDiscovery itself is an infinite stream of node ids you can't make this a combinator - no way to shuffle an infinite stream.

I guess we could maybe make a builder for the common case where you have a finite set of node ids, where you would have shuffled as a combinator.

examples/transfer-collection.rs
rklaehn: The store can be used without an endpoint. There are many very valid use cases for this, e.g. you have a local store and want to do something without ever listening on the network.

I guess you could do a thing which combines a store and an endpoint, but then that is what the downloader *is*. Maybe there needs to be a thing that combines a store, endpoint and downloader.

I want to keep the downloader as a separate thing though since it has some internal state and there might be situations where you want to create/destroy them or even have multiple independent of the store.

rklaehn (commented):
All fine except for that one bug with the collection tag.

examples/transfer-collection.rs
rklaehn: The store doesn't have anything, not even an endpoint. The **downloader** has a connection pool.

rklaehn (approved):

PR #177, opened 2025-10-02
feat(MemStore)!: expose garbage collection

b5:
## Description

This was a smaller change than expected.
* There is already a `gc_smoke_mem` test, so we know GC works with a memstore
* `store::mem::Options` was already defined, just empty. I've added a public `gc: Option<GcConfig>` field
* expanded `MemStore::new` to call `MemStore::new_with_opts` to expose providing custom GC options

## Breaking Changes

None.

## Notes & open questions

I haven't actually ensured that `MemStore::new_with_opts` works & is publically exposed. Should add a test for that

## Change checklist

- [ ] Self-review.
- [ ] Tests if relevant.


rklaehn:
Looks fine other than fmt.

The public API change is so inconsequential that I am tempted to call this "no change"...

A public type has stopped implementing one or more auto traits. This can break downstream code that depends on the traits being implemented.

Options is no longer RefUnwindSafe, nobody cares...

PR #178, opened 2025-10-03
feat(tags api)!: return number of tags removed by delete operations

b5:
## Description

It's a huge footgun to silently return `Ok(())`, even when no tags are actually removed. Returning the number of tags removed. Also adds documentation calling out this behaviour.

## Breaking Changes

breaks the return value of the tags API

## Notes & open questions

Should we do this with rename, and other mutating operations?

## Change checklist

- [ ] Self-review.
- [ ] Tests if relevant.


src/store/gc.rs
```diff
@@ -0,0 +1,472 @@
+use std::{collections::HashSet, pin::Pin, sync::Arc};
```
rklaehn: This has nothing to do with the PR, right?

Copies gc from fs to store top level. Probably related to something else?

rklaehn (commented):
Good improvement.

Seems pretty straightforward except for the new gc.rs file that is probably? unrelated to this PR.

PR #187, opened 2025-10-21
feat: compile to wasm for browsers

Frando:
## Description

Depends on https://github.com/n0-computer/irpc/pull/54 (merged but not released)

This makes iroh-blobs compile on wasm, with memstore only though. I had started on this a while back but it got much easier now with the fs-store behind a feature flag. 

Example: 
https://github.com/n0-computer/iroh-examples/pull/139
Deployed at 
https://n0-computer.github.io/iroh-examples/pr/139/browser-blobs/index.html
## Change checklist

- [ ] Self-review.
- [ ] Tests if relevant.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -35,15 +35,16 @@ chrono = "0.4.39"
 nested_enum_utils = "0.2.1"
 ref-cast = "1.0.24"
 arrayvec = "0.7.6"
-iroh = { version = "0.94", default-features = false }
+iroh = { version = "0.95", default-features = false }
 self_cell = "1.1.0"
 genawaiter = { version = "0.99.1", features = ["futures03"] }
-iroh-base = "0.94"
-iroh-tickets = "0.1"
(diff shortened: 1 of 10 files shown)


Cargo.toml
```diff
@@ -64,8 +63,24 @@ iroh = { version = "0.94", features = ["discovery-local-network"]}
...
 [features]
 hide-proto-docs = []
 metrics = []
-default = ["hide-proto-docs", "fs-store"]
-fs-store = ["dep:redb", "dep:reflink-copy"]
+default = ["hide-proto-docs", "fs-store", "rpc"]
+fs-store = ["dep:redb", "dep:reflink-copy", "bao-tree/fs"]
+rpc = ["dep:quinn", "irpc/rpc", "irpc/quinn_endpoint_setup"]
+
+[[example]]
```
rklaehn, line 77: So we will still be able to run these examples without --features because fs-store is on by default?

Frando: Yes. This is just for `cargo check --no-default-features --examples`, which is part of CI

rklaehn (approved):

PR #194, opened 2025-11-11
fix(wasm): use n0_future for all time::{Instant, Duration} references

b5:
## Description

While this will currently compile to wasm, when running in the browser, any reference to `std::time` will explode with unsupported platform errors, this fixes that.

## Change checklist

- [ ] Self-review.
- [ ] Tests if relevant.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -19,7 +19,7 @@ derive_more = { version = "2.0.1", features = ["from", "try_from", "into", "debu
 futures-lite = "2.6.0"
 quinn = { package = "iroh-quinn", version = "0.14.0", optional = true }
 n0-future = "0.3.1"
-n0-snafu = "0.2.2"
+n0-snafu = { version = "0.2.2", git="https://github.com/n0-computer/n0-snafu.git", branch="b5/wasm" }
 range-collections = { version = "0.4.6", features = ["serde"] }
 smallvec = { version = "1", features = ["serde", "const_new"] }
 snafu = "0.8.5"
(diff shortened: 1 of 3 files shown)


src/api.rs
```diff
@@ -127,7 +127,6 @@ impl From<irpc::Error> for ExportBaoError {
             irpc::Error::OneshotRecv { source, .. } => OneshotRecvSnafu.into_error(source),
             irpc::Error::Send { source, .. } => SendSnafu.into_error(source),
             irpc::Error::Request { source, .. } => RequestSnafu.into_error(source),
-            #[cfg(feature = "rpc")]
```
Frando, line 130: this needs to stay

Frando: it's the reason for the CI fails. We might want to fix this in irpc because it's a bit footgunny: The `irpc::Error::Write` variant is only present if the `rpc` feature of irpc is enabled. But for now just leave in the `cfg` here.

Cargo.toml
```diff
@@ -18,8 +18,8 @@ bytes = { version = "1", features = ["serde"] }
 derive_more = { version = "2.0.1", features = ["from", "try_from", "into", "debug", "display", "deref", "deref_mut"] }
 futures-lite = "2.6.0"
 quinn = { package = "iroh-quinn", version = "0.14.0", optional = true }
-n0-future = "0.3.0"
-n0-snafu = "0.2.2"
+n0-future = "0.3.1"
+n0-snafu = { version = "0.2.2", git="https://github.com/n0-computer/n0-snafu.git", branch="b5/wasm" }
```
matheus23, line 22: Need to remember to replace this with an actual dependency before merging.

PR #218, opened 2026-04-12
fix: add overflow protection for protocol input validation

emrul:
## Summary
- Add overflow protection for RangeSpec offset calculations using `checked_add`
- Add HashSeq size limit to prevent unbounded memory allocation from malicious input

## Test plan
- [x] All tests pass


emrul:

PR #219, opened 2026-04-12
fix: harden file import against symlink traversal and TOCTOU race

emrul:
## Summary
- Resolve symlinks before importing files to prevent path traversal attacks
- Reduce the TOCTOU window between path validation and file open

## Test plan
- [x] All tests pass


emrul:

PR #221, opened 2026-04-12
fix: correct misleading MAX_MESSAGE_SIZE doc comment

emrul:
## Summary
- The doc comment said "100MiB" but the value is 1MiB (1024 * 1024 = 1,048,576 bytes)
- Fix the comment to match the actual limit

## Test plan
- [x] All tests pass


emrul:

PR #226, opened 2026-05-08
fix: avoid ConnectionPool deadlock under concurrent get_or_connect

cbenhagen:
## Description

More than `PER_CONNECTION_INBOX` (100) concurrent `get_or_connect` calls for one slow-to-connect peer (one whose `endpoint.connect` doesn't fail fast, e.g. a backgrounded mobile peer or a stale address that takes the full `connect_timeout`) can deadlock the pool against requests for every other peer. Process exit is the only recovery.

What broke:

- `Actor::handle_msg` dispatches each `RequestRef` via `conn_tx.send(msg).await`. When concurrent requests for one peer exceed the per-connection inbox capacity (100), the pool main loop blocks here.
- The per-connection actor's connect-error path then calls `owner.close(id).await`, which sends `ConnectionShutdown` into the also-full pool main inbox and blocks too. Both actors mutually deadlocked.
- The actor was polled cooperatively from the pool's `FuturesUnordered`. While the pool was wedged inside `handle_msg`, it never polled `tasks.next()`, so `connect_fut` and the wrapping `connect_timeout` never made progress at all.

What changed:

- Connection actors now run as independent `n0_future::task::spawn` tasks (held via `AbortOnDropHandle` in `connections`). `Actor::tasks: FuturesUnordered<…>` is removed.
- All three `owner.close(id).await` sites in `run_connection_actor` are gone; each replaced with `break`/`return` plus a best-effort `try_send(ConnectionShutdown)` for eager cleanup. When the pool inbox is full and the eager notification is dropped, `handle_msg` reaps closed entries before falling back to idle eviction so churn against unreachable peers can't accumulate stale entries against `max_connections`.
- Connect-failure path closes `rx` first, then drains buffered `RequestRef`s with the cause, then exits. The pool inbox is no longer awaited from a connection actor.
- `ConnectionPool::idle()` (`async`, returned `Result`) is replaced with `try_idle()` (`fn`, returns `()`). The connection actor used the previous `Err` return as a shutdown signal; that path is now covered by `AbortOnDropHandle` aborting the actor when its entry is dropped, with equivalent end behavior.

What this PR does **not** fix: the per-peer head-of-line stall on `conn_tx.send().await`. Concurrent requests for one slow peer still serialize through that send for up to one `connect_timeout` window; concurrent storms across `N` distinct slow peers can compound to roughly `N * connect_timeout` worst-case dispatch latency for unrelated peers. That stall is bounded; the deadlock was not. Follow-up issue tracking the dispatch redesign: #227 

Tests:

- `connection_pool_dead_peer_backlog_does_not_wedge`: 150 concurrent requests to an unreachable peer plus a reachable-peer probe; without the fix the probe hangs.
- `connection_pool_dead_peer_below_inbox_cap_is_unaffected`: boundary check at 50 concurrent requests, isolating the regression to the inbox-overflow path.

## Breaking Changes

None. Public API unchanged.

## Notes & open questions

Alternatives considered and not taken:

- Generation tokens on connection entries: pure complexity tax once the `FuturesUnordered`-based cleanup was removed; lazy `Err`-on-`send` plus `AbortOnDropHandle` cover the same ground.
- `try_send` on `conn_tx` in the dispatch path: would add a Backpressure error variant callers don't need to handle, in exchange for trading the bounded head-of-line stall for surfaced errors. Not worth it.
- Failure caching with a TTL to suppress repeated dials of an unreachable peer: out of scope for this PR; separate behavioural change, separate proposal.

## Change checklist

- [x] Self-review.
- [x] Tests if relevant.

### DIFF
--- a/src/util/connection_pool.rs
+++ b/src/util/connection_pool.rs
@@ -23,11 +23,7 @@ use iroh::{
     Endpoint, EndpointId,
 };
 use n0_error::{e, stack_error};
-use n0_future::{
-    future::{self},
-    time::Duration,
-    FuturesUnordered, MaybeFuture, Stream, StreamExt,
-};
+use n0_future::{task::AbortOnDropHandle, time::Duration, MaybeFuture, Stream, StreamExt};
 use tokio::sync::{
     mpsc::{self, error::SendError as TokioSendError},
     oneshot, Notify,
@@ -36,6 +32,16 @@ use tracing::{debug, error, info, trace};
 
 pub type OnConnected =
     Arc<dyn Fn(&Endpoint, &Connection) -> n0_future::future::Boxed<io::Result<()>> + Send + Sync>;
+
+/// Capacity of the per-`EndpointId` inbox between the pool main loop and a
+/// connection actor. The bug this module's regression test guards against
+/// pivots on this constant: concurrent dispatch to one peer beyond this
+/// many requests will cause `Actor::handle_msg`'s `conn_tx.send().await`
+/// to block (the pool main loop's head-of-line stall), bounded by one
+/// `connect_timeout` window. The test scales its concurrency relative to
+/// this constant so the regression keeps testing what it claims to if
+/// the value is ever changed.
+const PER_CONNECTION_INBOX: usize = 100;
 
 /// Configuration options for the connection pool
 #[derive(derive_more::Debug, Clone)]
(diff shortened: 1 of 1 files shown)


rklaehn:
Thanks for the PR.

I will take a closer look as soon as I can.

As we prepare to 1.0 the connection pool has moved to https://github.com/n0-computer/iroh-util. Once we have this merged there we can do an update release for iroh-blobs.

cbenhagen:
Closing in favor of https://github.com/n0-computer/iroh-util/pull/1

Issue #227, opened 2026-05-08
ConnectionPool: per-peer backlog blocks unrelated-peer dispatch (bounded head-of-line stall)


cbenhagen:
Tracking issue, follow-up to the deadlock PR in #226. Not a blocker. The fix bounds the wedge to one `connect_timeout` per slow peer, but the underlying dispatch pattern remains: the pool main loop is the only dispatcher for all peers, so each cleared backlog still costs every other peer one stall window.

`Actor::handle_msg` awaits `entry.tx.send(msg)` into a per-connection `mpsc(PER_CONNECTION_INBOX = 100)`. When concurrency for one slow-to-connect peer exceeds 100 (the `P_pending` shape from #226: a peer whose `endpoint.connect` doesn't fail fast), that send parks for up to one `connect_timeout` window. The connection actor is suspended in `connect_timeout` and isn't draining its inbox; the actor's eventual rx close is what unblocks the pool. While parked, no other peer's `RequestRef` is dispatched. Worst case across N distinct slow peers each over the cap: roughly `N × connect_timeout` dispatch latency for an unrelated request.

Hits realistic workloads where unreachable peers don't fail fast: iroh-docs against many providers behind unstable NAT, or a startup-burst against a stale candidate list.

Mitigations (all deferred from the deadlock PR):

- `try_send` on `conn_tx` in dispatch. Adds a `Backpressure` error variant for callers.
- Failure caching with TTL. Sidesteps dispatch for known-unreachable peers; needs an invalidation story.
- Per-peer concurrency cap at `get_or_connect` entry. Admission instead of absorption.
- Per-peer dispatch tasks. Largest restructure, cleanest end state.

cbenhagen:
Closing in favor of https://github.com/n0-computer/iroh-util/issues/2

PR #229, opened 2026-05-08
deps: Use connection pool from iroh-util

rklaehn:
## Description

Use the connection pool from its new location in iroh-util.

It is reexported at the same location as before, so this should not be a breaking change.

## Breaking Changes

None. The semver check failure is just semver not understanding reeports. But it should work as is for downstream crates.

## Change checklist

- [ ] Self-review.
- [ ] Tests if relevant.


cbenhagen:
Oh, guess I will move #226 and #227 to https://github.com/n0-computer/iroh-util/ then :)

rklaehn:
> Oh, guess I will move #226 and #227 to https://github.com/n0-computer/iroh-util/ then :)

Sorry about the complication and the slow response. We are super busy getting iroh 1.0 rc out of the door and have neglected the proto crates a bit in the last months. Once 1.0 is out we will give them some love.

The connection pool for example is generically useful, but since we have done so many iroh releases we have kept it in iroh-blobs to keep things simple. But once we got 1.0 it can live in a separate crate without us having to publish it all the time.

Issue #235, opened 2026-05-27
Calling garbage collection manually


2color:
I noticed FsStore seems to support garbage collection through GcConfig/background scheduling, but not through a public explicit “run GC now” method. Was that intentional? IF so, I'm curious what the design tradeoff was. 

I ask because I imagine that in some situations, GC can be costly, and you might want to run it explicitly while there's less disk IO

matheus23:
Thanks for filing it (as we requested). For anyone reading this: We spoke about this on discord. The reason we've left out a manual "run_gc()" function is because GC shouldn't be costly if it's run reasonably frequently. If that is not the case, then that's an issue.

If anyone sees GC getting in their way because of high cost, or has a legitimate use case where running GC manually is the most sensible thing to do, then please let us know.

rikettsie:
I ran into exactly this scenario a couple of weeks ago downloading big blobs: https://github.com/rikettsie/ringdrop/blob/main/src/core/node.rs#L476

With GC configured at a 30-second interval, GC was heavily racing with active downloads, collecting partial .data especially on large blobs. The download failed or produced a corrupt result.

As you see in the code, I worked around it by creating a `batch()` scope while holding a `temp_tag` for the duration of the download, so as to prevent GC from collecting the blob while it's still being written:

```rust
let blob_batch = self.store.blobs().batch().await?;
let _gc_guard = blob_batch
    .temp_tag(HashAndFormat { hash, format })
    .await?;
// ... download advances here; _gc_guard keeps the blob alive
```

So, my use case is preventing GC from running on a specific blob during a critical window. 

This worked, the `temp_tag` API definitely covers that. But I think it's not so obvious from the docs.

PR #236, opened 2026-05-29
docs: add temp_tag pattern for protecting long-running downloads

rikettsie:
References #235

## Description

Batch::temp_tag offers blob protection from GC during a long-running write or download, but it was a little easy to miss for lack of documentation. This PR adds doc comments explaining the pattern:

- `Blobs::batch()`: explains what a batch scope is
- `Batch struct`: full use-case explanation and practical rust `# Example`
- `Batch::temp_tag()`: explains the protection mechanism
- `GcConfig`: adds a cross-reference pointing readers to `Batch::temp_tag`

## Notes & open questions

I followed the existing `async fn example()` doc-test convention used elsewhere in the codebase.

The explicit `drop(_guard)` at the end of the example is intentional: in my opinion it makes visible the lifetime boundary to the reader. Without it, someone could miss that is the moment protection ends. That said, if the scope closes right there, `_guard` goes obviously out of scope and GC becomes free to collect. So, the suggested explicit `drop` is only meaningful when you want to release protection **before** the surrounding scope ends.

## Change checklist

- [x] Self-review.
- [ ] Tests if relevant. (none, docs-only here)
- [ ] All breaking changes documented. (none)

### DIFF
--- a/src/api/blobs.rs
+++ b/src/api/blobs.rs
@@ -91,7 +91,7 @@ impl Blobs {
         Self::ref_cast(sender)
     }
 
-    /// Create a new batch scope.
+    /// Creates a new batch scope.
     ///
     /// A batch holds open a set of temporary tags that protect blobs from
     /// garbage collection for as long as the batch is alive. When the batch is
@@ -583,6 +583,7 @@ impl<'a> BatchAddProgress<'a> {
 /// // ... perform long-running write or network download here ...
 ///
 /// drop(_guard);
+/// // Any un-upgraded blobs from the batch are now eligible for garbage collection
 /// # Ok(())
 /// # }
 /// ```
@@ -630,7 +631,7 @@ impl<'a> Batch<'a> {
         }))
     }
 
-    /// Pin a hash inside this batch scope, protecting it from garbage collection.
+    /// Pins a hash inside this batch scope, protecting it from garbage collection.
     ///
     /// The returned [`TempTag`] keeps the blob alive until either the tag or
     /// the batch is dropped. This is the recommended way to protect a blob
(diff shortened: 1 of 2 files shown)


src/api/blobs.rs
```diff
@@ -91,6 +91,19 @@ impl Blobs {
         Self::ref_cast(sender)
     }
 
+    /// Create a new batch scope.
```
matheus23: We use the [rust style guide](https://rust-lang.github.io/rfcs/1574-more-api-documentation-conventions.html#summary-sentence) for writing documentation, which prefers using third person singular for the summary line.
So basically just this:

```suggestion
    /// Creates a new batch scope.
```

src/api/blobs.rs
```diff
@@ -588,6 +630,12 @@ impl<'a> Batch<'a> {
         }))
     }
 
+    /// Pin a hash inside this batch scope, protecting it from garbage collection.
```
matheus23:
```suggestion
    /// Pins a hash inside this batch scope, protecting it from garbage collection.
```

matheus23 (approved):
Just small nits, otherwise looking good! :)

matheus23:
> The explicit `drop(_guard)` at the end of the example is intentional: in my opinion it makes visible the lifetime boundary to the reader. Without it, someone could miss that is the moment protection ends. That said, if the scope closes right there, `_guard` goes obviously out of scope and GC becomes free to collect. So, the suggested explicit `drop` is only meaningful when you want to release protection **before** the surrounding scope ends.

That example reads well with the explicit `drop`, I like that :)
You could make your intention with the explicit drop a little more obvious by adding a comment *after* the `drop` line saying sth like `// Any un-upgraded blobs from the batch are now eligible for garbage collection`

src/api/blobs.rs
rikettsie: Sure! :100:

matheus23 (approved):

PR #259, opened 2026-09-01
feat(api): make ApiClient public and add Store::local for custom backends

hboisgibault:
## Motivation

`Store` was only constructible from in-crate backends (`MemStore`, `FsStore`) or via `Store::connect` (remote). External crates could not back a `Store` with a custom implementation (e.g. OPFS in the browser/WASM, in-memory test backend) by serving `proto::Request` on a local `irpc` channel, because `ApiClient` and the local constructor were `pub(crate)`.

## Changes

- Make `ApiClient` (`irpc::Client<proto::Request>`) public.
- Add `Store::local(client: ApiClient)` public constructor: create a `Store` from a custom local rpc client.
- Keep `Store::from_sender` as `pub(crate)` thin wrapper over `Store::local` (no breaking change).

## Example

```rust
let (client, rx) = irpc::channel::local(...); // serve proto::Request with your backend
tokio::spawn(async move { /* handle rx with custom backend, e.g. OPFS */ });
let store = Store::local(client);
```

### DIFF
--- a/src/api.rs
+++ b/src/api.rs
@@ -30,11 +30,7 @@ pub mod tags;
 use crate::{api::proto::WaitIdleRequest, provider::events::ProgressError};
 pub use crate::{store::util::Tag, util::temp_tag::TempTag};
 
-/// The local rpc client type backing an in-process [`Store`].
-///
-/// Public so that external crates can implement custom store backends:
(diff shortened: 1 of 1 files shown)


LegNeato:
Can we get this merged?

